PR #4, opened 2016-01-13
Check that the choosen page-dir exists before SNARE is launched. The …

nsmfoo:
…current function is to start SNARE but the "user" is unable to see the page (as it does not exist).

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -214,11 +214,15 @@ if __name__ == '__main__':
     parser.add_argument("--tanner", help="ip of the tanner service", default='tanner.mushmush.org')
     parser.add_argument("--skip-check-version", help="skip check for update", action='store_true')
     args = parser.parse_args()
-    future = loop.create_server(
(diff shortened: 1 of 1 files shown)


glaslos:
Can you use 4 spaces for indentation?

snare.py
```diff
@@ -214,11 +214,15 @@ def _run_cmd(cmd):
     parser.add_argument("--tanner", help="ip of the tanner service", default='tanner.mushmush.org')
     parser.add_argument("--skip-check-version", help="skip check for update", action='store_true')
     args = parser.parse_args()
-    future = loop.create_server(
-        lambda: HttpRequestHandler(args, debug=args.debug, keep_alive=75),
-        args.interface, args.port)
-    srv = loop.run_until_complete(future)
-    print("""
+    if not os.path.exists('/opt/snare/pages/' + args.page_dir):
+     print("--page-dir: " + args.page_dir + " does not exist")
```
glaslos:
```Python
print("--page-dir: {0} does not exist".format(args.page_dir))
```

PR #16, opened 2016-06-01
Config and web-tools

afeena:
Starting with config file and google/bing web-tools

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -25,6 +25,7 @@ import pwd
 import grp
 from urllib.parse import urlparse, unquote, parse_qsl
 import uuid
+import configparser
 
 import aiohttp
 from aiohttp.web import StaticRoute
@@ -201,6 +202,12 @@ def snare_setup():
     # Write pid to pid file
     with open('/opt/snare/snare.pid', 'wb') as pid_fh:
         pid_fh.write(str(os.getpid()).encode('utf-8'))
+
+    # config file
+    cfg_file = os.getcwd() + '/snare.cfg'
+    if not os.path.exists(cfg_file):
+        create_initial_config()
+
     # Read or create the sensor id
     uuid_file_path = '/opt/snare/snare.uuid'
     if os.path.exists(uuid_file_path):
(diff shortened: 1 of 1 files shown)


snare.py
```diff
@@ -201,6 +202,12 @@ def snare_setup():
     # Write pid to pid file
     with open('/opt/snare/snare.pid', 'wb') as pid_fh:
         pid_fh.write(str(os.getpid()).encode('utf-8'))
+
+    # config file
+    cfg_file = os.getcwd() + '/snare.cfg'
```
glaslos: Put the config file into `/opt/snare/`

snare.py
```diff
@@ -214,6 +221,13 @@ def snare_setup():
         return snare_uuid
 
 
+def create_initial_config():
```
glaslos: function definition should be before function call in line 209 (readability).

snare.py
```diff
@@ -214,6 +221,13 @@ def snare_setup():
...
+    config = configparser.ConfigParser()
```
glaslos: I think you can move this into the module scope (after the `if __name__ == '__main__'` and parsing the `--config` argument) so we can use it anywhere.

snare.py
```diff
@@ -190,6 +190,12 @@ def handle_request(self, request, payload):
         yield from response.write_eof()
 
 
+def create_initial_config():
+    config['WEB-TOOLS'] = dict(google='', bing='')
+    with open('snare.cfg', 'w') as configfile:
```
glaslos: This should be `/opt/snare/snare.cfg`

snare.py
```diff
@@ -319,6 +314,9 @@ def _run_cmd(cmd):
     parser.add_argument("--slurp-host", help="nsq logging host", default='slurp.mushmush.org')
     parser.add_argument("--slurp-auth", help="nsq logging auth", default='slurp')
     parser.add_argument("--config", help="snare config file", default='snare.cfg')
+
+    config = configparser.ConfigParser()
```
glaslos: Idea was to also read the config here. So either use the argument `--config` or the default path: `/opt/snare/snare.cfg`

glaslos:
Looks good now!

PR #22, opened 2016-06-14
Feature/submit error

afeena:
submit the data to TANNER even if error occurs

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -79,13 +79,14 @@ class HttpRequestHandler(aiohttp.server.ServerHttpProtocol):
         except Exception as e:
             print(e)
 
-    def create_data(self, request):
+    def create_data(self, request, response_status):
         data = dict(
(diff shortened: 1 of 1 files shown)


snare.py
```diff
@@ -79,6 +79,28 @@ def submit_slurp(self, data):
         except Exception as e:
             print(e)
 
+    def create_data(self, request):
+        data = dict(
```
glaslos: Do you think it would make sense to add the HTTP response status? Would make it easier for us to see if we were able to handle the request.

afeena: We can do it when sending errors, when we send normal requests, SNARE response depends on TANNER response., I can add response status field and make it None when sending normal request data

glaslos: I would set it to the HTTP code. So 200 for normal and 500 for error.

PR #24, opened 2016-06-29
Using pythongit to auto-update

afeena:


glaslos:
You have to update the requirements.txt file. Do you think we can bring this to the next level and check for updates every x hours?

afeena:
I think we can :) How often?

glaslos:
Make it a parameter with a sensible default value :)

afeena:
After some attempts and failures I have: 

- If we have loop, or any repeatable code for updates in snare, it can't update after first time because of snare drop privileges and repo can't be update:
```
git.exc.GitCommandError: 'git fetch -v origin' returned with exit code 255 
stderr: 'error: cannot open .git/FETCH_HEAD: Permission denied
```
- If we will have separate file for updates and run it as a subprocess, we need some kind of communication and the possibility to restart snare script from root if we have updates. 

Can you please give me advice how it would be better to organize updates?

glaslos:
Have you looked into forking the process so you have one with elevated permissions?

afeena:
I thought about it. Thanks, I'll try it :)

glaslos:
Well done!

PR #27, opened 2016-07-27
clone with replacing links in php

afeena:
Make possible to send post requests in login forms (on wp-login page, for example)

### DIFF
--- a/clone.py
+++ b/clone.py
@@ -18,7 +18,7 @@ import re
 import os
 import sys
 
-from urllib.parse import urlparse
+from urllib.parse import urlparse, unquote
 
 import asyncio
 import argparse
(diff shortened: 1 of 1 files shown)


clone.py
```diff
@@ -91,12 +101,13 @@ def get_body(self, root_url, urls, visited_urls):
         if re.match(patt, file_name):
             file_path, file_name = file_name.rsplit('/', 1)
             file_path += '/'
-        if parsed_url.query:
-            file_name += '?' + parsed_url.query
+            # if parsed_url.query:
+            # file_name += '?' + parsed_url.query
```
glaslos: You can remove those lines :)

afeena: Ooops, forgot to delete, thanks

PR #32, opened 2016-11-04
Updated imports to be compatible with aiohttp 1.1.0

wackyapples:
aiohttp 1.1.0 replaces StaticRoute with StaticResource (see [changelog](https://github.com/KeepSafe/aiohttp/blob/master/CHANGES.rst)).

Wrapped the new import in a try-except and alias StaticResource as StaticRoute to preserve compatibility with pervious versions of aiohttp.

I tested this with aiohttp 1.1.1 and 1.0.5, and both work.

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -29,8 +29,11 @@ import configparser
 import git
 import multiprocessing
 import aiohttp
-from aiohttp.web import StaticRoute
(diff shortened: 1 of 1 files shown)


glaslos:
Thanks for the fix!

PR #51, opened 2017-02-14
Fixes/content type fixes

### LINKED ISSUE #50: Mime types
Problems with css may occur because of the wrong mimetype
```
Python 3.5.2 (default, Dec 20 2016, 17:53:11)
[GCC 6.2.1 20160822] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import magic
>>> import mimetypes
>>> mimetypes.guess_type("style.css")
('text/css', None)
>>> magic.from_file("style.css",mime=True)
'text/plain'
>>>
```
nslwe0x 1

afeena:
#50 
https://github.com/ahupp/python-magic/issues/75

I think we should return mimetypes. It is stable and works well with css and html. If some types are missing - we can simply add them.

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -14,25 +14,24 @@ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
 """
 
-import os
-import sys
 import argparse
-import json
 import asyncio
-import pwd
-import grp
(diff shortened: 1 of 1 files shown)


glaslos:
Looks good and seems like a reasonable choice.

PR #52, opened 2017-02-16
remove path/query join

X4mp:
fixed an issue where files could not be served if requested with a url parameter


afeena:
Sometimes links can have the form such as ``foo.com/?p=1`` (e.g. on wp sites). In this case query part makes sense. It should be fixed in the cloner.

I've started some initial work [here ](https://github.com/afeena/snare/commit/075d4e8cce608abcd4260d77381fea6edb0f8d42) , but it should be tested carefully.

X4mp:
reopen for review

afeena:
We can keep these changes until we fix cloner

glaslos:
@afeena feel free to merge when you are happy with the PR

Issue #39, opened 2016-12-05
Clone depth


afeena:
I think we need parameter, which will restrict cloning depth

afeena:
#53

PR #53, opened 2017-03-04
Feature/cloner improvements

afeena:
I implemented new experimental cloner, which clones all the paths from the site. It replaces path and query with hash and stores it in a one directory. Also new cloner stores all metadata in json: source name, hash and content type. Because of the fact, that we get content type from the headers, we don't need to guess it (and make errors). In the future we can store more meta info, if we need it. 

**Why hashes?** 
Because sometimes we need to store path and query, especially when we had link such as foo.com/?p=21 (e.g. wordpress) Some queries quite long, so we can get os error. Hashes allows us store all types of the file. 

**Problems:**
- It gets all the files. All. So, if the site has a lot of pages, it can takes a long time. So we need to restrict cloning depth #39 
- Doesn't work with subdomains. 
- Maybe something else :) 

I tested it on some sites and it worked well, but more testing is required.

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -19,6 +19,7 @@ import asyncio
 import configparser
 import grp
 import json
+import mimetypes
 import multiprocessing
 import os
 import pwd
@@ -26,11 +27,10 @@ import sys
 import time
 import uuid
 from concurrent.futures import ProcessPoolExecutor
-from urllib.parse import urlparse, unquote, parse_qsl
+from urllib.parse import parse_qsl
 
 import aiohttp
 import git
-import mimetypes
 import pip
 from aiohttp import MultiDict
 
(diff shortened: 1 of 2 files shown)


snare.py
```diff
@@ -208,30 +215,24 @@ def handle_request(self, request, payload):
...
+            try:
+                file_name = self.meta[requested_name]['hash']
+                content_type = self.meta[requested_name]['content_type']
+            except KeyError:
+
                 response = aiohttp.Response(
                     self.writer, status=404, http_version=request.version
                 )
+            path = os.path.normpath(os.path.join(base_path, parsed_url))
+                if os.path.isfile(path) and path.startswith(self.dir):
```
X4mp: this causes an intentation error for me

snare.py
```diff
@@ -208,30 +215,24 @@ def handle_request(self, request, payload):
...
+                requested_name = self.run_args.index_page
+            try:
+                file_name = self.meta[requested_name]['hash']
+                content_type = self.meta[requested_name]['content_type']
+            except KeyError:
+
                 response = aiohttp.Response(
                     self.writer, status=404, http_version=request.version
                 )
+            path = os.path.normpath(os.path.join(base_path, parsed_url))
```
X4mp: parsed_url is not defined any more

X4mp: ok, I guess this is also due to some intentation?

afeena: I rebased the branch yesterday, I think it is the artefacts after that

afeena:
On this state there are some founded problems:

- Cloning files with another domain (on example.com, e.g.), despite the checking
- Links with mailto scheme
- Redirect tag don't have "value" in all cases

The list will be added and problems will be fixed as far as possible

afeena:
**Updates:**
1. Limit max depth of the cloner
2. Fix main problems

Need more testing! :)

rnehra01:
Using this cloner, the path `localhost:8080/` serves the `index.html` page while `localhost:8080/index.html` returns `404. Is it intentional? I think the latter too need to serve `index.html`.

afeena:
@rnehra01 Thanks! It's not expected behavior. All paths was stored in the meta file as `/foobar`, but `index.html` was stored without slash. That's why path `/index.html` was not found. Fixed

PR #54, opened 2017-03-05
Fix #6 

### LINKED ISSUE #6: Server header
We should either not provide a server header or use one provided by the user.

rnehra01:
Add an argument to take input `Server-header` and set its default value to `'None'` so that the server sends `'None'` if no input is provided

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -236,6 +236,7 @@ class HttpRequestHandler(aiohttp.server.ServerHttpProtocol):
                 response = aiohttp.Response(
                     self.writer, status=404, http_version=request.version
                 )
+        response.add_header('Server', self.run_args.server_header)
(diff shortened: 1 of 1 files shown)


afeena:
Why do you choose None as a default value? Maybe it is more reasonable to send something very common as a default not to capture the attention? (e.g. nginx). I'm not an expert, but if I were a bad guy, I would be very interested about None server value

rnehra01:
Okay, sounds good to me, I'll add nginx as default.

afeena:
Looks good, thanks for the pr :)

glaslos:
@afeena feel free to merge when you are happy with the PR

afeena:
@glaslos Ok, thanks :)

PR #62, opened 2017-03-22
Replace existing stale cookie

### LINKED ISSUE #60: Problem in cookie based session tracking
When we request snare with a random sess_uuid cookie already present, then tanner generates a new sess_uuid cookie with each request and old cookie doesn't get replaced.
Another way of saying it, if session gets expired, then the sess_uuid don't get replaced.

rnehra01:
When we request snare with a random sess_uuid cookie, then tanner generates a new sess_uuid cookie with each request and old cookie doesn't get replaced.
But this fix, old cookie will be replaced with the new one.

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -234,13 +234,17 @@ class HttpRequestHandler(aiohttp.server.ServerHttpProtocol):
                     self.writer, status=404, http_version=request.version
                 )
         response.add_header('Server', self.run_args.server_header)
+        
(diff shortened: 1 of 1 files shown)


glaslos:
Can you please give a brief description in the PR title and description?

rnehra01:
@glaslos done

glaslos:
I fixed the title as well.

PR #68, opened 2017-08-13
Snare-Tanner Communicaton

rnehra01:
Fix https://github.com/mushorg/tanner/issues/100

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -192,30 +192,9 @@ class HttpRequestHandler(aiohttp.server.ServerHttpProtocol):
         )
         content_type = None
         content = None
-        
-        if 'payload' in event_result['response']['message']['detection']:
-            payload_content = event_result['response']['message']['detection']['payload']
-            if type(payload_content) == dict:
-                try:
-                    file_name = self.meta[payload_content['page']]['hash']
-                    content_type = self.meta[payload_content['page']]['content_type']
-                    page_path = os.path.join(self.dir, file_name)
-                    with open(page_path, encoding='utf-8') as p:
-                        content = p.read()
-                except KeyError:
-                    content = '<html><body></body></html>'
-                    content_type = 'text\html'
-
-                soup = BeautifulSoup(content, 'html.parser')
-                script_tag = soup.new_tag('div')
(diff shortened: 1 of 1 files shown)


snare.py
```diff
@@ -235,6 +214,37 @@ def handle_request(self, request, payload):
...
+            else:
+                content_type = mimetypes.guess_type(payload_content['value'])[0]
+                content = payload_content['value'].encode('utf-8')
+
+            if 'headers' in payload_content:
+                for name, val in payload_content['headers']:
+                    response.add_header(name, val)
+        else:
+            response = aiohttp.Response(
+                self.writer, status=payload_content['status_code'], http_version=request.version
```
afeena: I think if we have some problem in tanner, we should keep it private and show e.g. index page for the user, how do you think?

snare.py
```diff
@@ -192,30 +192,9 @@ def handle_request(self, request, payload):
...
-                soup.body.append(script_tag)
-                content = str(soup).encode()
-
-            else:
-                content_type = mimetypes.guess_type(payload_content)[0]
-                content = payload_content.encode('utf-8')
-        else:
+       
+        detection = event_result['response']['message']['detection']
+        if detection['type'] == 1:
```
afeena: Maybe it is suitable to make separate function for parsing tanner response?

snare.py
```diff
@@ -257,6 +216,60 @@ def handle_request(self, request, payload):
...
+                soup.body.append(script_tag)
+                content = str(soup).encode()
+            else:
+                content_type = mimetypes.guess_type(payload_content['value'])[0]
+                content = payload_content['value'].encode('utf-8')
+
+            if 'headers' in payload_content:
+                headers =  payload_content['headers']
+        else:
+            status_code = payload_content['status_code']
```
afeena: If we have e.g. 500 error on tanner side, why do we need to pass it to the end user? Maybe it's better to return current page?

afeena: Ok, my bad :)
But what about tanner errors?

afeena:
I think we can merge it in develop :) I want to test it with tanner a bit

snare.py
rnehra01: Sure, I'll add that too.

PR #70, opened 2017-10-04
Pages converter

afeena:
Simple converter from old-style pages to new cloner style to prevent problems after auto-updates

### DIFF
--- a/converter.py
+++ b/converter.py
@@ -11,12 +11,13 @@ class Converter:
         self.meta = {}
 
     def convert(self, path):
-        f = []
+        files_to_convert = []
 
         for (dirpath, dirnames, filenames) in walk(path):
             for fn in filenames:
-                f.append(os.path.join(dirpath, fn))
-        for fn in f:
+                files_to_convert.append(os.path.join(dirpath, fn))
+
+        for fn in files_to_convert:
             path_len = len(path)
             file_name = fn[path_len:]
             m = hashlib.md5()
@@ -25,5 +26,6 @@ class Converter:
             self.meta[file_name] = {'hash': hash_name, 'content_type': mimetypes.guess_type(file_name)[0]}
             shutil.copyfile(fn, os.path.join(path, hash_name))
             os.remove(fn)
+
         with open(os.path.join(path, 'meta.json'), 'w')  as mj:
             json.dump(self.meta, mj)
--- a/converter.py
+++ b/converter.py
@@ -1,0 +1,29 @@
+import os
+import hashlib
+from os import walk
+import mimetypes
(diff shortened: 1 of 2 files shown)


converter.py
```diff
@@ -0,0 +1,29 @@
...
+import json
+import shutil
+
+
+class Converter:
+    def __init__(self):
+        self.meta = {}
+
+    def convert(self, path):
+        f = []
```
rnehra01: Can we give it a better name, maybe `files` ?

afeena: Sure :)

converter.py
```diff
@@ -0,0 +1,29 @@
...
+
+        for (dirpath, dirnames, filenames) in walk(path):
+            for fn in filenames:
+                f.append(os.path.join(dirpath, fn))
+        for fn in f:
+            path_len = len(path)
+            file_name = fn[path_len:]
```
rnehra01: Previously in line #L18, we can store only `filesnames` and here create a variable `full_file_name` or `full_path` by joining `file_name` and `path`. This method looks cleaner just the same you did in `sanre.py`

afeena: We can't store just `filenames`. Because we need full path like `/foo/bar/page.html` for hashing, not only `page`. `path` contains path like `/opt/snare/pages/<page>` and we need to ignore it.

rnehra01: Isn't `/foo/bar` = `/opt/snare/pages/example.com` = path? I guess that's why you're using `path_len` to determine `file_name` because `fn = path + file_name`.

afeena: No, this solution  based on `walk` result. Maybe there is a better solution :)

```
>>> import os
>>> for (dirpath, dirnames, filenames) in os.walk('/opt/snare/pages/example.com'):
...     for fn in filenames:
...             print(os.path.join(dirpath, fn))
...
/opt/snare/pages/example.com/page.html
/opt/snare/pages/example.com/foo/page.html
```

So we need `/foo/page.html` for hashing, not `/opt/snare/pages/example.com/foo/page.html` or `page.html`.  

`walk` by iteration
```
 for (dirpath, dirnames, filenames) in os.walk('/opt/snare/pages/example.com'):
...     print("dirpath: {}, dirnames: {}, filenames: {}".format(dirpath, dirnames, filenames))
...
dirpath: /opt/snare/pages/example.com, dirnames: ['foo'], filenames: ['page.html']
dirpath: /opt/snare/pages/example.com/foo, dirnames: [], filenames: ['page.html']
```

rnehra01: I didn't consider old style has directories too, my bad.

Issue #55, opened 2017-03-06
Deprecated Low-level Server


afeena:
We use Low-Level Server, which has been deprecated since aiohttp version 1.2. It needs to be rewritten asap.

http://aiohttp.readthedocs.io/en/stable/server.html

afeena:
aiohttp 2.0 was released yesterday (20.03), low-level server was deprecated.

glaslos:
@afeena what is the priority of this issue?

afeena:
@glaslos New server is already implemented in the develop branch, after testing we will release it.

afeena:
done in v0.3

PR #76, opened 2017-11-29
Access event logging to file

### LINKED ISSUE #73: Log timestamp and source IP addresses
The log should at least include timestamp, source IP address and the URI.

surajt97:
Log of timestamp, source ip and URL in /opt/snare/snare.log

### DIFF
--- a/logger.py
+++ b/logger.py
@@ -14,15 +14,16 @@ class LevelFilter(logging.Filter):
 
 class Logger:
     @staticmethod
-    def create_logger(debug_filename, logger_name):
+    def create_logger(debug_filename, logger_name, level_name):
         logger = logging.getLogger(logger_name)
-        logger.setLevel(logging.INFO)
+        level_dict = {'CRITICAL': 50, 'ERROR': 40, 'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
+        logger.setLevel(level_dict[level_name])
         logger.propagate = False
         formatter = logging.Formatter(fmt='%(hostIP)s - %(user)s [%(time)s]%(message)s\"%(req)s\" %(stat)d %(content_length)s')
 
-        # DEBUG log to '/opt/snare/snare.log'
+        # Log to '/opt/snare/snare.log'
         debug_log_handler = logging.handlers.RotatingFileHandler(debug_filename, encoding='utf-8')
-        debug_log_handler.setLevel(logging.INFO)
+        debug_log_handler.setLevel(level_dict[level_name])
         debug_log_handler.setFormatter(formatter)
         max_level_filter = LevelFilter(logging.ERROR)
         debug_log_handler.addFilter(max_level_filter)
--- a/logger.py
+++ b/logger.py
@@ -1,5 +1,6 @@
 import logging
 import logging.handlers
+from time import gmtime
 
 
 class LevelFilter(logging.Filter):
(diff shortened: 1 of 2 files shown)


glaslos:
Please make this optional using a command line flag.

snare.py
```diff
@@ -178,6 +179,10 @@ def handle_request(self, request, payload):
         # Submit the event to the TANNER service
         event_result = yield from self.submit_data(data)
 
+        #Log the URL to /opt/snare/snare.log
+        ip=self.transport.get_extra_info('peername')[0]
+        self.logger.info('Source IP %s - - Requested path %s', ip, self.run_args.page_dir + request.path)
```
glaslos: Please use the Apache access log format: https://httpd.apache.org/docs/current/logs.html#accesslog

snare.py
```diff
@@ -461,6 +466,11 @@ def check_tanner_connection():
         args.interface, int(args.port))
     srv = loop.run_until_complete(future)
 
+    #log info
+    info_log_file_name = '/opt/snare/snare.log'
+    logger.Logger.create_logger(info_log_file_name, __package__)
```
glaslos: Pass the logger to `HttpRequestHandler` as argument.

logger.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def filter(self, record):
+        return record.levelno < self.level # "<" instead of "<=": since logger.setLevel is inclusive, this should be exclusive
+
+
+class Logger:
+    @staticmethod
+    def create_logger(debug_filename, logger_name):
+        logger = logging.getLogger(logger_name)
+        logger.setLevel(logging.INFO)
```
glaslos: Expose the log level as command line flag

logger.py
```diff
@@ -0,0 +1,30 @@
...
+
+        # DEBUG log to '/opt/snare/snare.log'
+        debug_log_handler = logging.handlers.RotatingFileHandler(debug_filename, encoding='utf-8')
+        debug_log_handler.setLevel(logging.INFO)
+        debug_log_handler.setFormatter(formatter)
+        max_level_filter = LevelFilter(logging.ERROR)
+        debug_log_handler.addFilter(max_level_filter)
+        logger.addHandler(debug_log_handler)
+
+        return logger
```
glaslos: missing newline

snare.py
```diff
@@ -178,6 +179,10 @@ def handle_request(self, request, payload):
         # Submit the event to the TANNER service
         event_result = yield from self.submit_data(data)
 
+        #Log the URL to /opt/snare/snare.log
+        ip=self.transport.get_extra_info('peername')[0]
```
glaslos: PEP-8

glaslos:
@surajt97 I consider the formatting of the messages as critical. If we stick to the same format as Apache, people can use other existing access log tools to analyze the Snare logs.

surajt97:
@glaslos should I use aiohttp.access logger?

glaslos:
Yes, that could be an option. Details can be found here: https://docs.aiohttp.org/en/stable/logging.html
You can also just change the format of your current logger. Using aiohttp might require fixing this issue #55 and using something more high level.

glaslos:
@afeena do you have an opinion on this?

afeena:
I'd prefer using aiohttp logging, it can be done even without high-level changes, but solution without aiohttp logging is also ok for me:)

snare.py
```diff
@@ -251,6 +256,19 @@ def handle_request(self, request, payload):
...
+        req = request.method + ' /' + self.run_args.page_dir + request.path
+        major = response.version.major
+        minor = response.version.minor
+        ver = " HTTP/" + str(major) + '.' + str(minor)
+        req = req + ver
+        user = getpass.getuser()
+        tim = strftime("%d/%b/%Y:%H:%M:%S %z")
+        d = {'hostIP': request.headers['Host'], 'stat': response.status, 'req': req, 'user': user,
+             'content_length': response.headers['Content-Length'], 'time': tim}
+        self.logger.info(' ', extra=d)
```
afeena: I think this part with logging should executed only if we have `logging==True`. In other case this line will be executed with default self.logger of aiohttp server

surajt97: @afeena pls review.

afeena:
@surajt97  Could you please also take a look at these comments? https://github.com/mushorg/snare/pull/76#discussion_r153746461

glaslos:
Needs a rebase

mzfr:
@afeena I think we should close this since this issue is fixed after merging #109

Issue #79, opened 2017-12-31
Migrating to latest syntax (python 3.5) for coroutines


viskey98:
Currently snare is using the 'asyncio.coroutine/yield from' syntax. But ever since python 3.5 we have 'async/await' syntax which is better than the former. We can migrate snare to the new syntax.

PR #80, opened 2018-01-03
Migrating to the latest syntax for coroutines

viskey98:
Currently snare is using the 'asyncio.coroutine/yield from' syntax. But ever since python 3.5 we have 'async/await' syntax which is better than the former. We can migrate snare to the new syntax.

### DIFF
--- a/snare.py
+++ b/snare.py
@@ -378,7 +378,7 @@ def parse_timeout(timeout):
     return result
 
 
-async def check_tanner_connection():
+async def check_tanner():
     vm = VersionManager()
     with aiohttp.ClientSession() as client:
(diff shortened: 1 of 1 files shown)


viskey98:
This PR fixes issue #79

afeena:
I don't mind using new syntax, but it will break the compatibility with python <3.5 and because of auto-updates it can cause a problem for snares, which are working now. On the other hand, we are on our early-stage so it may be reasonable to migrate right now. 
I'd like to get more opinions about this question :)

viskey98:
Yah , we can keep this for discussion and when we feel its actually necessary we can migrate it :)

viskey98:
new pr for develop branch #91

PR #81, opened 2018-01-20
Logging for snare (#73)

### LINKED ISSUE #73: Log timestamp and source IP addresses

mzfr:
Fixes #73 

I added a logger kind of similar to tanner.

### DIFF
--- a/logger.py
+++ b/logger.py
@@ -1,0 +1,36 @@
+import logging
+import logging.handlers
+
+class LevelFilter(logging.Filter):
+    def __init__(self, level):
+        self.level = level
+
+    def filter(self, record):
+        return record.levelno < self.level
+
+class Logger:
+    @staticmethod
+    def create_logger(debug_filename, err_filename, logger_name):
+
+
+        logger = logging.getLogger(logger_name)
+        logger.setLevel(logging.DEBUG)
+        logger.propagate = False
+        formatter = logging.Formatter(fmt='%(asctime)s %(levelname)s:%(name)s:%(funcName)s: %(message)s',
+                                      datefmt='%Y-%m-%d %H:%M')
+
+        # ERROR log to 'snare.err'
+        error_log_handler = logging.handlers.RotatingFileHandler(err_filename, encoding='utf-8')
+        error_log_handler.setLevel(logging.ERROR)
+        error_log_handler.setFormatter(formatter)
+        logger.addHandler(error_log_handler)
+
+        # DEBUG log to 'snare.log'
+        debug_log_handler = logging.handlers.RotatingFileHandler(debug_filename, encoding='utf-8')
+        debug_log_handler.setLevel(logging.DEBUG)
+        debug_log_handler.setFormatter(formatter)
+        max_level_filter = LevelFilter(logging.ERROR)
+        debug_log_handler.addFilter(max_level_filter)
+        logger.addHandler(debug_log_handler)
+
+        return logger
--- a/logger.py
+++ b/logger.py
@@ -1,30 +1,22 @@
 import logging
 import logging.handlers
 
-class LevelFilter(logging.Filter):
-    def __init__(self, level):
-        self.level = level
-
-    def filter(self, record):
-        return record.levelno < self.level
 
 class Logger:
+
     @staticmethod
     def create_logger(log_filename, logger_name, level_name):
         logger = logging.getLogger(logger_name)
-        logging_levels = {'CRITICAL': 50, 'ERROR': 40, 'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
+        logging_levels = {'CRITICAL': 50, 'ERROR': 40,
+                          'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
         logger.setLevel(logging_levels[level_name])
-        logger.propagate = False
         formatter = logging.Formatter(fmt='%(asctime)s %(levelname)s:%(name)s:%(funcName)s: %(message)s',
                                       datefmt='%Y-%m-%d %H:%M:%S')
-
-
         # Log to 'snare.log'
-        log_handler = logging.handlers.RotatingFileHandler(log_filename, encoding='utf-8')
+        log_handler = logging.handlers.RotatingFileHandler(
+            log_filename, encoding='utf-8')
         log_handler.setLevel(logging_levels[level_name])
         log_handler.setFormatter(formatter)
-        max_level_filter = LevelFilter(logging.ERROR)
-        log_handler.addFilter(max_level_filter)
         logger.addHandler(log_handler)
 
         return logger
--- a/logger.py
+++ b/logger.py
@@ -7,15 +7,12 @@ class Logger:
     @staticmethod
     def create_logger(log_filename, logger_name, level_name):
         logger = logging.getLogger(logger_name)
-        logging_levels = {'CRITICAL': 50, 'ERROR': 40,
-                          'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
-        logger.setLevel(logging_levels[level_name])
+        logger.setLevel(level_name)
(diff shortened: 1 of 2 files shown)


glaslos:
Please rebase and add a description.

mzfr:
@glaslos I made changes :)

logger.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, level):
+        self.level = level
+
+    def filter(self, record):
+        return record.levelno < self.level
+
+class Logger:
+    @staticmethod
+    def create_logger(debug_filename, err_filename, logger_name):
+
```
glaslos: Empty space?

logger.py
```diff
@@ -0,0 +1,36 @@
...
+
+        logger = logging.getLogger(logger_name)
+        logger.setLevel(logging.DEBUG)
```
glaslos: Please expose the log level

logger.py
```diff
@@ -0,0 +1,36 @@
...
+        logger.propagate = False
+        formatter = logging.Formatter(fmt='%(asctime)s %(levelname)s:%(name)s:%(funcName)s: %(message)s',
+                                      datefmt='%Y-%m-%d %H:%M')
```
glaslos: Add seconds

logger.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # ERROR log to 'snare.err'
+        error_log_handler = logging.handlers.RotatingFileHandler(err_filename, encoding='utf-8')
+        error_log_handler.setLevel(logging.ERROR)
+        error_log_handler.setFormatter(formatter)
+        logger.addHandler(error_log_handler)
+
+        # DEBUG log to 'snare.log'
+        debug_log_handler = logging.handlers.RotatingFileHandler(debug_filename, encoding='utf-8')
+        debug_log_handler.setLevel(logging.DEBUG)
```
glaslos: You first set the level to DEBUG and then you filter out everything below ERROR?

logger.py
```diff
@@ -10,27 +10,21 @@ def filter(self, record):
 
 class Logger:
     @staticmethod
-    def create_logger(debug_filename, err_filename, logger_name):
-
-
+    def create_logger(log_filename, logger_name, level_name):
         logger = logging.getLogger(logger_name)
-        logger.setLevel(logging.DEBUG)
+        logging_levels = {'CRITICAL': 50, 'ERROR': 40, 'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
```
glaslos: you don't need to pass integers, the name of the level is fine.

logger.py
```diff
@@ -0,0 +1,30 @@
...
+class Logger:
+    @staticmethod
+    def create_logger(log_filename, logger_name, level_name):
+        logger = logging.getLogger(logger_name)
+        logging_levels = {'CRITICAL': 50, 'ERROR': 40, 'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
+        logger.setLevel(logging_levels[level_name])
+        logger.propagate = False
+        formatter = logging.Formatter(fmt='%(asctime)s %(levelname)s:%(name)s:%(funcName)s: %(message)s',
+                                      datefmt='%Y-%m-%d %H:%M:%S')
+
```
glaslos: two spaces

logger.py
```diff
@@ -0,0 +1,30 @@
...
+
+        # Log to 'snare.log'
+        log_handler = logging.handlers.RotatingFileHandler(log_filename, encoding='utf-8')
+        log_handler.setLevel(logging_levels[level_name])
```
glaslos: first you set the level here and then you overwrite it in line 26 and then you add a filter in line 27

snare.py
```diff
@@ -445,12 +450,16 @@ def parse_timeout(timeout):
...
+    parser.add_argument("--logger", help="log of time-stamp, URL and source IP")
     args = parser.parse_args()
     base_path = '/opt/snare/'
     base_page_path = '/opt/snare/pages/'
     config = configparser.ConfigParser()
     config.read(os.path.join(base_path,args.config))
 
+    if args.logger:
+        log_file_name = "/opt/snare/snare.log"
+        logger.Logger.create_logger(log_file_name, __package__,args.logger)
```
glaslos: missing space, read PEP8

snare.py
```diff
@@ -445,12 +450,16 @@ def parse_timeout(timeout):
     parser.add_argument("--auto-update", help="auto update SNARE if new version available ", default=True)
     parser.add_argument("--update-timeout", help="update snare every timeout ", default='24H')
     parser.add_argument("--server-header", help="set server-header", default='nginx')
+    parser.add_argument("--logger", help="log of time-stamp, URL and source IP")
```
glaslos: what arguments am I supposed to pass to the logger flag?

snare.py
```diff
@@ -445,12 +450,16 @@ def parse_timeout(timeout):
...
     args = parser.parse_args()
     base_path = '/opt/snare/'
     base_page_path = '/opt/snare/pages/'
     config = configparser.ConfigParser()
     config.read(os.path.join(base_path,args.config))
 
+    if args.logger:
+        log_file_name = "/opt/snare/snare.log"
```
glaslos: please let me use an argument to change the location.

logger.py
```diff
@@ -0,0 +1,30 @@
...
+    def filter(self, record):
+        return record.levelno < self.level
+
+class Logger:
+    @staticmethod
+    def create_logger(log_filename, logger_name, level_name):
+        logger = logging.getLogger(logger_name)
+        logging_levels = {'CRITICAL': 50, 'ERROR': 40, 'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
+        logger.setLevel(logging_levels[level_name])
+        logger.propagate = False
```
glaslos: can you explain why you set this to false?

mzfr: I think setting it to true will attach the handler with logger and it will log same thing twice

mzfr:
I didn't remove those integers because I think it make setting log level easy otherwise  we'll have to use something like [this](https://docs.python.org/2/library/logging.config.html) which will make logger a bit complex

glaslos:
you think `logger.setLevel("DEBUG")` doesn't work?

logger.py
```diff
@@ -0,0 +1,22 @@
...
+        logger = logging.getLogger(logger_name)
+        logging_levels = {'CRITICAL': 50, 'ERROR': 40,
+                          'WARNING': 30, 'INFO': 20, 'DEBUG': 10, 'NOTSET': 0}
+        logger.setLevel(logging_levels[level_name])
+        formatter = logging.Formatter(fmt='%(asctime)s %(levelname)s:%(name)s:%(funcName)s: %(message)s',
+                                      datefmt='%Y-%m-%d %H:%M:%S')
+        # Log to 'snare.log'
+        log_handler = logging.handlers.RotatingFileHandler(
+            log_filename, encoding='utf-8')
+        log_handler.setLevel(logging_levels[level_name])
```
glaslos: why do you set the level here?

snare.py
```diff
@@ -429,29 +442,53 @@ def parse_timeout(timeout):
...
     args = parser.parse_args()
     base_path = '/opt/snare/'
     base_page_path = '/opt/snare/pages/'
     config = configparser.ConfigParser()
-    config.read(os.path.join(base_path,args.config))
-
+    config.read(os.path.join(base_path, args.config))
 
+    if args.logger_level:
+        if args.log_file:
```
glaslos: this is always true with the default set

mzfr: This was fixed

mzfr:
@glaslos Should I move this PR to develop branch?

afeena:
@mzfr thanks for your work :)
Please look at the #109  - I merged this because it contains actual logging into the file. If you have any improvements - feel free to suggest :)

Issue #83, opened 2018-01-21
can't start snare.py


perl2php:
hello
i'm really new to python
installed snare as described
system is ubuntu 16.04, python3 3.5.2
when trying to start:

ubuntu@test:~/snare$ sudo python3 snare.py --port 8080 --page-dir example.com
Traceback (most recent call last):
  File "snare.py", line 37, in <module>
    from aiohttp.web import StaticResource as StaticRoute
  File "/home/ubuntu/.local/lib/python3.5/site-packages/aiohttp/web.py", line 10, in <module>
    from . import (hdrs, web_exceptions, web_middlewares, web_reqrep, web_server,
  File "/home/ubuntu/.local/lib/python3.5/site-packages/aiohttp/web_middlewares.py", line 5, in <module>
    from aiohttp.web_urldispatcher import SystemRoute
  File "/home/ubuntu/.local/lib/python3.5/site-packages/aiohttp/web_urldispatcher.py", line 17, in <module>
    from yarl import URL, unquote
ImportError: cannot import name 'unquote'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "snare.py", line 39, in <module>
    from aiohttp.web import StaticResource
  File "/home/ubuntu/.local/lib/python3.5/site-packages/aiohttp/web.py", line 10, in <module>
    from . import (hdrs, web_exceptions, web_middlewares, web_reqrep, web_server,
  File "/home/ubuntu/.local/lib/python3.5/site-packages/aiohttp/web_middlewares.py", line 5, in <module>
    from aiohttp.web_urldispatcher import SystemRoute
  File "/home/ubuntu/.local/lib/python3.5/site-packages/aiohttp/web_urldispatcher.py", line 17, in <module>
    from yarl import URL, unquote
ImportError: cannot import name 'unquote'

what should i do?

afeena:
Hi, please run 
```pip3 freeze``` 
and show the output

Seems like you have problems with yarl version 
https://github.com/aio-libs/aiohttp/issues/2662

perl2php:
ubuntu@test:~$  pip3 freeze
aiohttp==1.3.5
async-timeout==2.0.0
beautifulsoup4==4.6.0
blinker==1.3
chardet==3.0.4
cloud-init==17.1
command-not-found==0.3
configobj==5.0.6
cryptography==1.2.3
cssutils==1.0.2
gitdb2==2.0.3
GitPython==2.1.8
idna==2.6
[...]

PR #85, opened 2018-01-23
Update requirements.txt

mzfr:
With yarl==1.1.0 , SNARE would give a traceback error as mentioned in #83

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,4 @@ cssutils
 gitpython
 netifaces
 yarl==0.9.8
-<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -3,5 +3,9 @@ beautifulsoup4
 cssutils
 gitpython
 netifaces
-yarl
-python-magic
\ No newline at end of file
+yarl==0.9.8
+<<<<<<< HEAD
```
afeena: please delete these lines. They are artifacts from git :)

Issue #66, opened 2017-06-20
Custom error page


eliuha:
When trying to access to not existing page, snare shows generic 404 error, which is replaces by chrome to standard error message 

`This 127.0.0.2 page can’t be found
No webpage was found for the web address: http://127.0.0.2:8080/aaa/wp-content/aaa
Search Google for 127 8080 aaa content
HTTP ERROR 404
`

wouldn't it make sense to redirect to the index page or show a custom error?

afeena:
Hi @elihua, thanks for the issue :) It will be good feature for the snare :) I think we could try to fetch error page style from the original site while cloning

rnehra01:
Also, when we access a path like `/some_path?foo=bar` and if no attack is present in `foo` param then snare return 404(assume some_path exists). Usually, most websites return `/some_path` if we pass arbitrary GET params to it. So we can include that too.

glaslos:
Solved by #93

PR #87, opened 2018-01-30
Custom error page

mzfr:
Updated clone.py to fetch error page style from original site.(fixes issue #66 )

_Not quite ready yet_

### DIFF
--- a/clone.py
+++ b/clone.py
@@ -147,8 +147,8 @@ class Cloner(object):
             try:
                 with aiohttp.Timeout(10.0):
                     with aiohttp.ClientSession() as session:
-                        response = yield from session.get(current_url, headers={'Accept': 'text/html'})
-                        data = yield from response.read()
+                        response = await session.get(current_url, headers={'Accept': 'text/html'})
+                        data = await response.read()
             except aiohttp.ClientError as client_error:
 
                     response = await session.get(current_url)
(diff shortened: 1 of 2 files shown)


viskey98:
Hi , @mzfr your 'custom error page' commit doesn't work and it has already been tested by a similar code done by me , it doesn't fetch the custom error page of the website , but a file stating "not found". And Iam currently working on that issue and i have completed it and currently discussing with @afeena about it, so u can let this issue (#66) aside and focus on the other ones :)

mzfr:
@viskey98 Can you post the error that you get? I've tried it on a bunch of pages and it seems to work just fine.

viskey98:
@mzfr  The commit you have made will just show a web-page with "NOT FOUND" written on it , when we request a page which is not found . But ,what we expect is : to show up a web-page with a custom 404 error page in the theme of the website which is cloned , for example look at this page (http://github.com/error_page) with the picture in the background and all the customised theme.

mzfr:
@viskey98 That is very site-specific. Some sites try to be smart about the content they serve based on the headers they receive. In this particular case, Github uses the `Accept` header to send different content.

Running this command should give 'NOT FOUND':
``` curl 'https://github.com/error_page' ```

While running this will give the correct HTML page:
``` curl 'https://github.com/error_page' -H 'Accept: text/html' ```

I've pushed a fix for this!

File: clone.py [hunk 1]
```diff
@@ -145,6 +146,11 @@ def _make_filename(self, url):
             content_type = None
             try:
                 with aiohttp.Timeout(10.0):
+                    with aiohttp.ClientSession() as session:
+                        response = yield from session.get(current_url, headers={'Accept': 'text/html'})
```
viskey98: use await syntax

clone.py
```diff
@@ -145,6 +146,11 @@ def _make_filename(self, url):
...
+                        data = yield from response.read()
```
viskey98: Please correct ur code by changing all "yield from" syntax into "await syntax"

clone.py
```diff
@@ -169,6 +175,17 @@ def _make_filename(self, url):
...
+                            yield from self.new_urls.put(carved_url)
+
+    @asyncio.coroutine
+    def run(self):
+        yield from self.new_urls.put(self.root)
+        # Force 404 Page
+        yield from self.new_urls.put(self.error_page)
+        return (yield from self.get_body())
+
                         if carved_url.human_repr() not in self.visited_urls:
```
viskey98: After the release , the whole code of this pr has messed up , please try to run this code and fix the errors :)

glaslos:
Solved by #93

Issue #90, opened 2018-01-30
Error after latest commit on snare.py develop branch


viskey98:
Following error occurs : 
   _____ _   _____    ____  ______
  / ___// | / /   |  / __ \/ ____/
  \__ \/  |/ / /| | / /_/ / __/
 ___/ / /|  / ___ |/ _, _/ /___
/____/_/ |_/_/  |_/_/ |_/_____/

    
Traceback (most recent call last):
  File "snare.py", line 491, in <module>
    args.host-ip, int(args.port))
AttributeError: 'Namespace' object has no attribute 'host'

PR #91, opened 2018-01-30
Fix for error after latest commit on snare.py develop branch

viskey98:
Fixes issue #90


viskey98:
@afeena as we are going to make a new release , I feel it would be nice if we migrate our syntax of snare.py from (ayncio.couroutine / yield from) to the latest (async / await) syntax which is latest and better.
(Keeping in mind clone.py already has the latest syntax but not snare.py).

Issue #77, opened 2017-11-29
Add Test/Travis


glaslos:
Add some tests and run them in Travis on commit.

glaslos:
Added snare to travis. Added the badge to the README.

PR #96, opened 2018-02-03
Tests

mzfr:
This is an attempt for the issue #77 
I have added test only for versions_manager and I am working on tests for others.

Let me know if there is some changes required.


versions_manager.py
```diff
@@ -13,5 +13,5 @@ def check_compatibility(self, tanner_version):
         min_version = self.version_mapper[self.version][0]
         max_version = self.version_mapper[self.version][1]
         if not (StrictVersion(min_version) <= StrictVersion(tanner_version) <= StrictVersion(max_version)):
-            print("Wrong tanner version: {}. Compatible versions are {} - {}".format(tanner_version, min_version, max_version))
+            raise RuntimeError("Wrong tanner version: {}. Compatible versions are {} - {}".format(tanner_version, min_version, max_version))
```
afeena: if you use RuntimeError, you don't need exit. But you can also assert with SystemExit in test if you use exit()

afeena:
Nice! I am really happy that snare will have tests :)

PR #106, opened 2018-02-14
Added the feature to disable dorks insertion

Aws0mus:
I just want to deploy the web without adding extra code.


glaslos:
Can you explain your motivation?

Aws0mus:
If you deploy a simple website (e.g. a login form) as a honeypot and contains some dorks about php or other technologies, I believe it is easy to detect as a honeypot. I would like to be able to disable them when necessary.

snare.py
```diff
@@ -135,32 +135,33 @@ def create_data(self, request, response_status):
...
-                if len(self.dorks) <= 0:
-                    self.dorks = await self.get_dorks()
-                word += ' '
-                if idx % 5 == 0:
-                    a_tag = soup.new_tag(
-                        'a',
-                        href=self.dorks.pop(),
-                        style='color:{color};text-decoration:none;cursor:text;'.format(
-                            color=css.color if css and 'color' in css.keys() else '#000000'
+        if self.run_args.no_dorks != True:
```
glaslos: How about we move the whole dork insertion code into a separate function?

afeena: I think we can add it as an issue and work on it as a separate task :)

PR #117, opened 2018-03-25
Format code according to pep8 style and add pep8 check in travis-ci

mzfr:
This will check for the pep8 violation.
Also I have fixed all the errors that were there in code base so now only the newly changed code will be checked.


snare.py
```diff
@@ -73,7 +74,7 @@ def __init__(self, meta, run_args, debug=False, keep_alive=75, **kwargs):
                         print(e)
                     finally:
                         r.release()
-        except:
+        except Exception:
```
afeena: it will be really good, if we change Exception here to specific aiohttp exception. Because using broader Exception is very bad :)

snare.py
```diff
@@ -463,15 +462,16 @@ def parse_timeout(timeout):
...
+
     if not os.path.exists(os.path.join(full_page_path, 'meta.json')):
         conv = Converter()
         conv.convert(full_page_path)
         print("pages was converted. Try to clone again for the better result.")
 
     with open(os.path.join(full_page_path, 'meta.json')) as meta:
         meta_info = json.load(meta)
-    if not os.path.exists(os.path.join(base_page_path,args.page_dir,os.path.join(meta_info[args.index_page]['hash']))):
+    if not os.path.exists
```
afeena: this line is not working because of wrapping
```
  File "snare.py", line 473
    if not os.path.exists
                        ^
SyntaxError: invalid syntax

```

mzfr: @afeena  I have resolved the issue. Please take a look

mzfr: @afeena I have fixed the issue. Please take a look.

afeena: @mzfr Are you sure you commit your changes? It seems like the same problem still occurs

mzfr:
@afeena I am sorry I actually thought I have fixed the problem while rebase-ing.
Now it is done.

Issue #113, opened 2018-03-12
Clone error on CSSUTILS


viskey98:
I tried cloning my Institute's Intranet website, and faced many errors in CSS utils.
I captured the errors in an error log and this is what i found:
```shell

2018-03-12 12:52:49 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -webkit-gradient(linear, left top, left bottom, color-stop(0%, #0272a7), color-stop(100%, #013953)) [44:2: background]
2018-03-12 12:52:49 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -webkit-linear-gradient(top, #FFF 0%, #FFF 100%) [45:2: background]
2018-03-12 12:52:49 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -o-linear-gradient(top, #FFF 0%, #FFF 100%) [46:2: background]
2018-03-12 12:52:49 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -ms-linear-gradient(top, #FFF 0%, #FFF 100%) [47:2: background]
2018-03-12 12:52:49 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: linear-gradient(top, #FFF 0%, #FFF 100%) [48:2: background]
2018-03-12 12:52:49 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: 16x [350:2: font-size]
2018-03-12 12:52:49 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -webkit-gradient(linear, 0% 0%, 0% 100%, from(#EEE), to(#BBB)) [423:2: background]
2018-03-12 12:52:50 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -webkit-gradient(linear, left top, left bottom, from(#1490ac), to(#006c82)) [113:2: background]
2018-03-12 12:52:50 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -webkit-linear-gradient(top, #1490ac, #006c82) [114:2: background]
2018-03-12 12:52:50 ERROR:CSSUTILS:__handle: Property: Invalid value for "CSS Level 2.1" property: -moz-linear-gradient(top, #1490ac, #006c82) [115:2: background]
[...]

PR #120, opened 2018-04-01
Add new parameter css-validate to cloner

viskey98:
This will create a new parameter `css-validate` to set whether css validation is required or not.
Fixes issue #113

### DIFF
--- a/clone.py
+++ b/clone.py
@@ -32,7 +32,7 @@ import logging
 
 
 class Cloner(object):
-    def __init__(self, root, max_depth):
+    def __init__(self, root, max_depth, css_validate):
         self.visited_urls = []
         self.root, self.error_page  = self.add_scheme(root)
         self.max_depth = max_depth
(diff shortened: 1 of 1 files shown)


clone.py
```diff
@@ -165,8 +165,11 @@ def _make_filename(self, url):
                     data = str(soup).encode()
                 with open(os.path.join(self.target_path, hash_name), 'wb') as index_fh:
                     index_fh.write(data)
-                if content_type == 'text/css':
-                    css = cssutils.parseString(data)
+                if content_type == 'text/css':                       
+                    if self.css_validate == "False":
+                        css = cssutils.parseString(data, validate=False)
```
afeena: Maybe it's better to write:
```cssutils.parseString(data, validate=self.validate)```
In this case you don't need this if/else statement

viskey98:
@afeena Keeping your suggestion in mind.
As the argparse gets input as string (even if we use `type=bool`, any input will validate to True as it is direct string to bool conversion). 
According to me there are two solutions for that:
1. Create a custom function `str_to_bool()` (in a desired way) and in argparse use `type=str_to_bool` .
2. Or else we have to take input as a string in argparse (using `type=str, default="True"`). And have to use an if-else statement before passing the parameter to the Cloner class or after. 
Now, I have changed the code and implemented the first one, please suggest any changes if required :)

afeena:
@viskey98 good work :)

viskey98:
@afeena Thank you :)

PR #125, opened 2018-05-14
Test converter

viskey98:
This is the unittest for testing converter by checking if the hashed files created by converter actually exist or not.
Please suggest any changes if required :)

### DIFF
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -19,17 +19,14 @@ class TestConverter(unittest.TestCase):
         self.hname2 = ""
         with open(os.path.join(self.page_path, 'index.html'),   'w') as f:
             f.write(self.content)
-            f.close()
         with open(os.path.join(self.page_path, 'depth/page.html'), 'w') as f:
             f.write(self.content)
-            f.close()
         self.cnv = Converter()
 
     def test_converter(self):
         self.cnv.convert(self.page_path)
-        f = open(os.path.join(self.page_path, 'meta.json'), 'r')
-        s = f.read()
-        f.close()
+        with open(os.path.join(self.page_path, 'meta.json'), 'r') as f:
+            s = f.read()
         index = s.index('"index.html"') + len('"index.html"') + 12
         self.hname1 = s[index: index + 32]
         index = s.index('"depth/page.html"') + len('"depth/page.html"') + 12
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -9,7 +9,7 @@ from converter import Converter
 class TestConverter(unittest.TestCase):
 
     def setUp(self):
-        self.content = '<html><body>sample</body></html>'
+        self.content = '<html><body> </body></html>'
         self.page_path = '/tmp/test/'
         if not os.path.exists('/tmp/test'):
             os.mkdir('/tmp/test')
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -23,7 +23,7 @@ class TestConverter(unittest.TestCase):
     def test_converter(self):
         self.cnv.convert(self.page_path)
         with open(os.path.join(self.page_path, 'meta.json')) as f:
-            s = json.loads(f.read())
+            s = json.load(f)
         self.hname1 = s['index.html']['hash']
         self.hname2 = s['depth/page.html']['hash']
         assert(os.path.exists(self.page_path + self.hname1) and
--- a/tests/test_converter.py
+++ b/tests/test_converter.py
@@ -2,14 +2,13 @@ import unittest
 import os
 import sys
 import shutil
-import hashlib
 from converter import Converter
 
 
 class TestConverter(unittest.TestCase):
 
     def setUp(self):
-        self.content = '<html><body></body></html>'
+        self.content = '<html><body>sample</body></html>'
         self.page_path = '/tmp/test/'
         if not os.path.exists('/tmp/test'):
             os.mkdir('/tmp/test')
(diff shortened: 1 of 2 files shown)


tests/test_converter.py
```diff
@@ -0,0 +1,42 @@
...
+        self.page_path = '/tmp/test/'
+        if not os.path.exists('/tmp/test'):
+            os.mkdir('/tmp/test')
+        if not os.path.exists('/tmp/test/depth'):
+            os.mkdir('/tmp/test/depth')
+        self.hname1 = ""
+        self.hname2 = ""
+        with open(os.path.join(self.page_path, 'index.html'),   'w') as f:
+            f.write(self.content)
+            f.close()
```
afeena: no need to use `close` with `with`

tests/test_converter.py
```diff
@@ -0,0 +1,42 @@
...
+        with open(os.path.join(self.page_path, 'depth/page.html'), 'w') as f:
+            f.write(self.content)
+            f.close()
+        self.cnv = Converter()
+
+    def test_converter(self):
+        self.cnv.convert(self.page_path)
+        f = open(os.path.join(self.page_path, 'meta.json'), 'r')
```
afeena: it's better to use `with` while reading the file

viskey98:
@afeena The suggested changes have been made. This test runs fine in my machine but travis sometimes passes and sometimes fails for the same code. Why does that occur?

tests/test_converter.py
```diff
@@ -0,0 +1,38 @@
...
+        with open(os.path.join(self.page_path, 'index.html'),   'w') as f:
+            f.write(self.content)
+        with open(os.path.join(self.page_path, 'depth/page.html'), 'w') as f:
+            f.write(self.content)
+        self.cnv = Converter()
+
+    def test_converter(self):
+        self.cnv.convert(self.page_path)
+        with open(os.path.join(self.page_path, 'meta.json'), 'r') as f:
+            s = f.read()
```
afeena: why don't you use `json.load` here?

tests/test_converter.py
```diff
@@ -0,0 +1,38 @@
...
+        index = s.index('"index.html"') + len('"index.html"') + 12
```
afeena: parse file as json and work with it

tests/test_converter.py
```diff
@@ -0,0 +1,38 @@
...
+        self.hname2 = ""
+        with open(os.path.join(self.page_path, 'index.html'),   'w') as f:
+            f.write(self.content)
+        with open(os.path.join(self.page_path, 'depth/page.html'), 'w') as f:
+            f.write(self.content)
+        self.cnv = Converter()
+
+    def test_converter(self):
+        self.cnv.convert(self.page_path)
+        with open(os.path.join(self.page_path, 'meta.json'), 'r') as f:
```
afeena: 'r' is default property, no need to specify

tests/test_converter.py
```diff
@@ -0,0 +1,38 @@
...
+            s = f.read()
+        index = s.index('"index.html"') + len('"index.html"') + 12
+        self.hname1 = s[index: index + 32]
+        index = s.index('"depth/page.html"') + len('"depth/page.html"') + 12
```
afeena: same here

tests/test_converter.py
```diff
@@ -0,0 +1,38 @@
...
+import shutil
+from converter import Converter
+
+
+class TestConverter(unittest.TestCase):
+
+    def setUp(self):
+        self.content = '<html><body></body></html>'
+        self.page_path = '/tmp/test/'
+        if not os.path.exists('/tmp/test'):
```
afeena: for creating multiple directories you can use `os.makedirs`

tests/test_converter.py
```diff
@@ -0,0 +1,38 @@
...
+        index = s.index('"index.html"') + len('"index.html"') + 12
+        self.hname1 = s[index: index + 32]
+        index = s.index('"depth/page.html"') + len('"depth/page.html"') + 12
+        self.hname2 = s[index: index + 32]
+        assert(os.path.exists(self.page_path + self.hname1) and
+               os.path.exists(self.page_path + self.hname2))
+
+    def tearDown(self):
+        shutil.rmtree('/tmp/test/depth')
+        shutil.rmtree('/tmp/test')
```
afeena: when you delete parent directory, others are also deleted. no need to delete it directly

viskey98:
@afeena The suggested changes have been done, please look into it :)

tests/test_converter.py
```diff
@@ -0,0 +1,34 @@
...
+            f.write(self.content)
+        with open(os.path.join(self.page_path, 'depth/page.html'), 'w') as f:
+            f.write(self.content)
+        self.cnv = Converter()
+
+    def test_converter(self):
+        self.cnv.convert(self.page_path)
+        with open(os.path.join(self.page_path, 'meta.json')) as f:
+            s = f.read()
+            s = json.loads(s)
```
afeena: json can be loaded from buffer directly

viskey98:
@afeena suggested change has been made :)

tests/test_converter.py
```diff
@@ -0,0 +1,33 @@
...
+        with open(os.path.join(self.page_path, 'index.html'), 'w') as f:
+            f.write(self.content)
+        with open(os.path.join(self.page_path, 'depth/page.html'), 'w') as f:
+            f.write(self.content)
+        self.cnv = Converter()
+
+    def test_converter(self):
+        self.cnv.convert(self.page_path)
+        with open(os.path.join(self.page_path, 'meta.json')) as f:
+            s = json.loads(f.read())
```
afeena: please use `json.load(f)` here :)

Json has two function: `load` and `loads`:
`load` can directly read from text stream
`loads` read json from a string

viskey98:
As I was dealing with a string I got a little confused there whether to use `load()` or `loads()`. Now I have implemented with load directly from the buffer. Please look into it

PR #127, opened 2018-05-15
Test get_dorks

viskey98:
This PR will test get_dorks by checking if the get_dorks() function (of snare.py) was able to connect to default tanner and get dorks.

### DIFF
--- a/tests/test_get_dorks.py
+++ b/tests/test_get_dorks.py
@@ -1,0 +1,35 @@
+import unittest
+import asyncio
+import argparse
+import aiohttp
+import shutil
+import os
+from snare import HttpRequestHandler
+
+
+class TestGetDorks(unittest.TestCase):
+    def setUp(self):
+        self.meta = {}
+        run_args = argparse.ArgumentParser();
+        run_args.add_argument("--tanner")
+        run_args.add_argument("--page-dir")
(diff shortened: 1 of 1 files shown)


afeena:
What are you actually testing here? If you want to test that you get exception when server is not available, use assertRaise https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

viskey98:
I am trying to make sure no exception occurs when I run that function. assertRaises will assert if the exception occurs and will pass the test for the same. But I wish to fail the test in case of any exception regarding the connection and pass otherwise. That is why I didn't use assertRaises.

viskey98:
I have re-written the test by implementing :
- test_get_dorks: testing if get_dorks returns a empty list '[]' without raising any exception.

- test_get_dorks_fail: testing if get_dorks fails with a non-existent server by raising a specified exception.

viskey98:
@afeena I have re-written the test (using mocks) by testing follwing behaviour:
- test_get_dorks: to check if the session.get() was called with correct parameters
- test_return_dorks: to check if dorks obtained from ClientSession.json() is returned properly.
- test_return_dorks_exception: to check if the exception occuring in the function is raised properly.

PR #130, opened 2018-05-17
Test handle_html_content

viskey98:
This will test the following behaviours of handle_html_content functionality of snare.py:
- test_handle_content: To check if the html content is handled properly and expected content is returned.
- test_handle_content_no_dorks: To check if the html content is handled properly without dorks and expected content is returned.
- test_handle_content_exception: To check if exception is raised properly from the function.


tests/test_handle_html_content.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class TestHandleHtmlContent(unittest.TestCase):
+    def setUp(self):
+        self.meta = {}
+        run_args = argparse.ArgumentParser()
+        run_args.add_argument("--tanner")
+        run_args.add_argument("--page-dir")
+        run_args.add_argument("--no-dorks")
+        if not os.path.exists("/opt/snare/pages/test"):
```
afeena: what happens if someone has directory target 'test"?

viskey98: If someone has directory target "test" this test will automatically delete it at the end of the test.
I think we should check the existence of the directory and leave it be if it already exists after the test. Or delete the folder if we created it in the test.

viskey98: Please check if it is okay now. :)

tests/test_add_meta_tag.py
```diff
@@ -10,7 +10,9 @@ class TestAddMetaTag(unittest.TestCase):
 
     def setUp(self):
         if not os.path.exists("/opt/snare/pages/test"):
+            self.path_exists = False
```
afeena: what is the purpose of this flag variable? if you have this directory, you will override index file, it might be crucial. And also it will be always True

I guess it's the problem that we can change default folder '/opt/snare/pages'. In this case, probably will be better to create folder with random character sequence and not just use test

and makedirs by default raises the exception, if file exists

tests/test_handle_html_content.py
```diff
@@ -0,0 +1,78 @@
...
+
+        async def test():
+            self.return_content = await self.handler.handle_html_content(self.content)
+        self.loop.run_until_complete(test())
+        soup = BeautifulSoup(self.return_content, "html.parser")
+        self.return_content = soup.decode("utf-8")
+        self.assertEquals(self.return_content, self.no_dorks_content)
+
+    def test_handle_content_exception(self):
+        self.handler = HttpRequestHandler(self.meta, self.args)
```
afeena: do you really need to create new handler for every test?

tests/test_add_meta_tag.py
```diff
@@ -10,7 +10,9 @@ class TestAddMetaTag(unittest.TestCase):
...
             os.makedirs("/opt/snare/pages/test")
+        self.path_exists = True
```
afeena: I guess it should be with the else branch

tests/test_add_meta_tag.py
afeena: Looking at the other prs I understand the purposes of this flag and it can be used as a temporal solution before we change our structure :)

PR #131, opened 2018-05-17
Test handle_error

viskey98:
This will test handle_error functionality of snare by checking the following behaviors:
- test_create_error_data: To check if the create_data() was called with the right parameters
- test_submit_error_data: To check if submit_data() was called with the right parameters
- test_handle_error_data: To check if super().handle_error() was called with right parameters


tests/test_handle_error.py
```diff
@@ -0,0 +1,86 @@
...
+from unittest.mock import Mock
+import asyncio
+import argparse
+import aiohttp
+import shutil
+import os
+from snare import HttpRequestHandler
+
+
+class AsyncMock(Mock):  # custom function defined to mock asyncio coroutines
```
afeena: if you are going to use this class multiple times, it's better to make a separate class (I would like to suggest you to make something like utils module)

tests/test_handle_error.py
```diff
@@ -0,0 +1,86 @@
...
+            headers="test_headers",
+            uuid="test_uuid",
+            peer="test_peer",
+            status="test_status",
+            error=self.exc
+        )
+        aiohttp.server.ServerHttpProtocol.handle_error = Mock()
+
+    def test_create_error_data(self):
+        self.handler = HttpRequestHandler(self.meta, self.args)
```
afeena: you have similar setup for handler in all functions - maybe it's better to move it into setup?

tests/test_add_meta_tag.py
```diff
@@ -10,7 +10,10 @@ class TestAddMetaTag(unittest.TestCase):
 
     def setUp(self):
         if not os.path.exists("/opt/snare/pages/test"):
+            self.path_exists = False
             os.makedirs("/opt/snare/pages/test")
+        else:
+            self.path_exists = True
         self.content = '<html><head>title</head><body>sample</body></html>'
         self.page_dir = "test"
         self.index_page = "index.html"
```
afeena: if folder test exists, you will override index.html

tests/test_add_meta_tag.py
```diff
@@ -4,17 +4,18 @@
...
 class TestAddMetaTag(unittest.TestCase):
 
     def setUp(self):
-        if not os.path.exists("/opt/snare/pages/test"):
-            os.makedirs("/opt/snare/pages/test")
+        self.main_page_path = generate_unique_path()
+        os.makedirs(self.main_page_path)
         self.content = '<html><head>title</head><body>sample</body></html>'
-        self.page_dir = "test"
+        self.page_dir = self.main_page_path[-9:]
```
afeena: you can use `rsplit` here. we should avoid "magic constants" or make it clear from context. here it is not obvious why it's 9

tests/test_add_meta_tag.py
```diff
@@ -4,17 +4,18 @@
...
+        print(self.main_page_path)
```
afeena: delete print

tests/test_get_dorks.py
```diff
@@ -28,19 +17,19 @@ def setUp(self):
         run_args = argparse.ArgumentParser()
         run_args.add_argument("--tanner")
         run_args.add_argument("--page-dir")
-        if not os.path.exists("/opt/snare/pages/test"):
-            os.makedirs("/opt/snare/pages/test")
+        self.main_page_path = generate_unique_path()
+        os.makedirs(self.main_page_path)
         self.args = run_args.parse_args(['--tanner', 'test'])
```
afeena: when you parse arguments step by step, you get this
```
>>> args = run_args.parse_args(['--tanner', 'bar'])
>>> args
Namespace(page_dir=None, tanner='bar')
>>> args = run_args.parse_args(['--page-dir', 'foo'])
>>> args
Namespace(page_dir='foo', tanner=None)
```

PR #133, opened 2018-05-19
Test create_data

viskey98:
This is to test create_data functionality of snare by checking the following behaviour:
- test_create_data: to check if expected data is created for the given input parameters.


tests/test_add_meta_tag.py
```diff
@@ -10,7 +10,10 @@ class TestAddMetaTag(unittest.TestCase):
 
     def setUp(self):
         if not os.path.exists("/opt/snare/pages/test"):
```
afeena: this part is completely similar for most of the tests. We can think about separating it.
And instead of test folder to avoid using flags we can generate random string, it also might have collisions, but the probability is near to 0

PR #135, opened 2018-05-20
Test parse_tanner_response

viskey98:
This will test the parse_tanner_response functionality of snare by testing the following behavoir:
- Test_parse_type_one: To check if type one request is parsed as expected.
- Test_parse_type_two: To check if type two request is parsed as expected.
- Test_parse_type_three: To check if  type three request is parsed as expected.
- Test_parse_exception: To check if exception is raised properly by the function.


tests/test_parse_tanner_response.py
```diff
@@ -0,0 +1,106 @@
...
+
+        async def test():
+            (self.res1, self.res2,
+             self.res3, self.res4) = await self.handler.parse_tanner_response(self.requested_name, self.detection)
+        self.loop.run_until_complete(test())
+        self.assertEquals(self.res1, self.page_content)
+        self.assertEquals(self.res2, self.content_type)
+        self.assertEquals(self.res3, {})
+        self.assertEquals(self.res4, 200)
+        self.handler.handle_html_content.assert_called_with(self.call_content)
```
afeena: here in one test you check different cases, it's better to test one case per test

tests/test_parse_tanner_response.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_parse_type_one(self):
+        self.detection = {"type": 1}
+        self.call_content = b'<html><body></body></html>'
+        self.expected_content = self.page_content
+
+        async def test():
+            (self.res1, self.res2,
+             self.res3, self.res4) = await self.handler.parse_tanner_response(self.requested_name, self.detection)
+        self.loop.run_until_complete(test())
+        self.assertEquals(self.res1, self.page_content)
```
afeena: can we compare parse result as a tuples? e.g. with this https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

viskey98:
@afeena I have done the changes suggested by you. Please look into it :)

tests/test_parse_tanner_response.py
```diff
@@ -0,0 +1,113 @@
...
+        self.handler.run_args.index_page = '/index.html'
+        self.handler.handle_html_content = AsyncMock(return_value=self.page_content)
+
+    def test_parse_type_one(self):
+        self.detection = {"type": 1}
+        self.call_content = b'<html><body></body></html>'
+        self.expected_content = self.page_content
+
+        async def test():
+            (self.res1, self.res2,
```
afeena: you don't need `self` for result variables here, you will not reuse these variables. So for the future - it's better to avoid such things

PR #136, opened 2018-05-25
Rewriting snare to latest aiohttp.web server

viskey98:
This is the first re-writing of snare to latest server.


viskey98:
There was a problem in this PR i will re-create it

PR #137, opened 2018-05-25
Re-writing snare with latest web server

viskey98:
This will re-write snare to latest web server .

### DIFF
--- a/middlewares.py
+++ b/middlewares.py
@@ -6,38 +6,29 @@ class SnareMiddleware():
 
     def __init__(self, file_name):
         self.error_404 = file_name
-    
+
     async def handle_404(self, request):
         return aiohttp_jinja2.render_template(self.error_404, request, {})
 
-
     async def handle_500(self, request):
         return aiohttp_jinja2.render_template('500.html', request, {})
-
 
     def create_error_middleware(self, overrides):
 
         @web.middleware
         async def error_middleware(request, handler):
-
             try:
                 response = await handler(request)
-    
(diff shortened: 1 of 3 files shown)


server.py
```diff
@@ -0,0 +1,276 @@
...
+                headers = payload_content['headers']
+        else:
+            payload_content = detection['payload']
+            status_code = payload_content['status_code']
+
+        return (content, content_type, headers, status_code)
+
+    def start(self):
+        app = web.Application()
+        app.add_routes([web.route('*', '/{tail:.*}', self.handle_request)])
```
afeena: could you please explain a bit this `/{tail:.*}` path?

viskey98: This is to route all http request to the handle_request handler. This is because our handler was written to handle all kind of http requests for all the paths of the specific web site (eg: '/', '/path1',  '/path/two') If we don't use this `/{tail:.*}` path, then we need to create a handler each for all the paths of all websites we are going to serve, which is practically impossible

viskey98: Check this link for more clarity :) 
https://stackoverflow.com/questions/34565705/asyncio-and-aiohttp-route-all-urls-paths-to-handler

middlewares.py
```diff
@@ -0,0 +1,47 @@
+import aiohttp_jinja2
```
afeena: please fix the formatting in this file

viskey98: This is fixed please look into it :)

viskey98:
Fixed by latest PR #141

PR #138, opened 2018-05-27
Breaking server.py 

viskey98:
This PR will break server.py. This was done with the base of latest rewritten server. After the re-written server gets merged, this pr will again be re-based if necessary. :)

### DIFF
--- a/server.py
+++ b/server.py
@@ -1,0 +1,276 @@
+import argparse
+import asyncio
+import configparser
+import grp
+import json
+import mimetypes
+import multiprocessing
+import os
+import pwd
+import sys
+import time
+import uuid
+from concurrent.futures import ProcessPoolExecutor
+from urllib.parse import urlparse, unquote, parse_qsl
+from versions_manager import VersionManager
+import aiohttp
+import git
+import pip
+from aiohttp import web
+import re
+import logging
+import logger
+import multidict
+import aiohttp_jinja2
+import jinja2
+
+try:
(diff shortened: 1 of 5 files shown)


server.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def start(self):
+        app = web.Application()
+        app.add_routes([web.route('*', '/{tail:.*}', self.handle_request)])
+        aiohttp_jinja2.setup(
+            app, loader=jinja2.FileSystemLoader(self.dir)
+        )
+        middleware = SnareMiddleware(self.meta['/status_404']['hash'])
+        middleware.setup_middlewares(app)
+        web.run_app(app, host=self.run_args.host_ip)
```
afeena: and what about port?

server.py
```diff
@@ -0,0 +1,106 @@
...
+        self.sroute = StaticRoute(
+            name=None, prefix='/',
+            directory=self.dir
+        )
+        self.tanner_handler = TannerHandler(run_args, meta, snare_uuid)
+
+    async def handle_request(self, request):
+        self.logger.info('Request path: {0}'.format(request.path))
+        data = self.tanner_handler.create_data(request, 200)
+        if request.method == 'POST':
```
afeena: maybe we can split post and get handlers?

viskey98: what about PUT and other requests? If we split both handlers, we need to define handlers for others HTTP request as well. What do you think about it?

viskey98:
Fixed by latest PR #141

PR #140, opened 2018-05-28
add Dockerfile, docker-compose.yml

t3chn0m4g3:
Build with:
`docker-compose build`

Run with:
`docker-compose up`

Cloning from remote to avoid including unnecessary files and keeping the resulting image slim. Opted against Docker multi-stage build which resulted in slightly larger image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+FROM alpine
+
+# Setup apt
+RUN apk -U --no-cache add \
+               build-base \
+               git \
+               linux-headers \
+               python3 \
+               python3-dev && \ 
+
+# Setup Snare
+    git clone --depth=1 https://github.com/mushorg/snare /opt/snare && \
+    cd /opt/snare/ && \
+    pip3 install --no-cache-dir --upgrade pip setuptools && \
+    pip3 install --no-cache-dir -r requirements.txt && \
+    python3.6 clone.py --target http://example.com && \
+    
+# Clean up
+    apk del --purge \
+            build-base \
+            linux-headers \
+            python3-dev && \
+    rm -rf /root/* && \
+    rm -rf /tmp/* /var/tmp/* && \
+    rm -rf /var/cache/apk/*
+
+# Start snare
+WORKDIR /opt/snare
+CMD /usr/bin/python3.6 /opt/snare/snare.py --no-dorks --auto-update false --host-ip 0.0.0.0 --port 80 --page-dir example.com
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,19 @@
+version: '2.3'
+
+networks:
+  snare_local:
+
+services:
+
+# Snare service
+  snare:
+    build: .
+    container_name: snare
+    restart: always
+    stop_signal: SIGKILL
+    tty: true
+    networks:
+     - snare_local
+    ports:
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+               linux-headers \
+               python3 \
+               python3-dev && \ 
+
+# Setup Snare
+    git clone --depth=1 https://github.com/mushorg/snare /opt/snare && \
+    cd /opt/snare/ && \
+    pip3 install --no-cache-dir --upgrade pip setuptools && \
+    pip3 install --no-cache-dir -r requirements.txt && \
+    python3.6 clone.py --target http://example.com && \
```
mainnika: It's better to use ARG here.
https://docs.docker.com/engine/reference/builder/#arg

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+            build-base \
+            linux-headers \
+            python3-dev && \
+    rm -rf /root/* && \
+    rm -rf /tmp/* /var/tmp/* && \
+    rm -rf /var/cache/apk/*
+
+# Start snare
+WORKDIR /opt/snare
+CMD /usr/bin/python3.6 /opt/snare/snare.py --no-dorks --auto-update false --host-ip 0.0.0.0 --port 80 --page-dir example.com
```
mainnika: You've forgotten EXPOSE

Dockerfile
t3chn0m4g3: In your opinion, what is the advantage of using ARGs here? You'd have to rebuild anyway using a `build-arg list`, hence writing it directly to the `Dockerfile` seemed the obvious choice.
In a matter of using variables one could argue to remove `line 16` and incorporate it in the `CMD` while setting ENVs in the `docker-compose.yml`:

```
CMD python3.6 clone.py --target $TARGET && exec /usr/bin/python3.6 /opt/snare/snare.py --no-dorks --auto-update false --host-ip 0.0.0.0 --port 80 --page-dir $TARGET_NAME
```

Dockerfile
t3chn0m4g3: I intentionally left it out and I have never used `EXPOSE` for my production `Dockerfiles`. It simply is not necessary (you use `ports` in `docker-compose.yml` or `docker run -p <src_port>:<dest_port>` for correct startup) and it even messes with things should a user decide to build an image which is based (`FROM`) on the one built with this `Dockerfile` (if published on a registry).

Dockerfile
mainnika: Good idea about ENV!
Then it would be better to check the $TARGET folder before.
I'd suggest something like this for the entrypoint script
```sh
#!/bin/sh

PYTHON=python3.6

if [ ! -d "$TARGET" ]; then
  $PYTHON clone.py --target $TARGET
fi

$PYTHON snare.py --no-dorks --auto-update false --host-ip 0.0.0.0 --port 80 --page-dir $TARGET

```

Dockerfile
mainnika: But it is better to use it in a common dockerfile, isn't it? 
If I use some custom logic based on EXPOSE. For example, nginx-proxy docker image is working with EXPOSE port to proxy it.

afeena:
Thanks for your pr :)

PR #153, opened 2018-07-28
Update docs 

viskey98:
This will update the docs and readme.md after the latest SNARE release.

### DIFF
--- a/docs/source/parameters.rst
+++ b/docs/source/parameters.rst
@@ -1,6 +1,6 @@
 Snare command line parameters
 =============================
-snare.py [**--page-dir** *folder* ] [**--list-pages**]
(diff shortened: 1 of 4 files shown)


docs/source/parameters.rst
```diff
@@ -1,6 +1,6 @@
 Snare command line parameters
 =============================
-snare.py [**--page-dir** *folder* ] [**--list-pages**]
+snare [**--page-dir** *folder* ] [**--list-pages**]
```
afeena: we have 1 param left here - host-ip, could you please add it?

viskey98: I have added it :)

Issue #154, opened 2018-08-02
KeyError 'status_404'


t3chn0m4g3:
When running latest Snare 0.3 I am getting the following error and Snare exits:

```
/opt/snare # snare --debug true --no-dorks true --auto-update false --host-ip 0.0.0.0 --port 80 --page-dir example.com

   _____ _   _____    ____  ______
  / ___// | / /   |  / __ \/ ____/
  \__ \/  |/ / /| | / /_/ / __/
 ___/ / /|  / ___ |/ _, _/ /___
/____/_/ |_/_/  |_/_/ |_/_____/

    
privileges dropped, running as "nobody:nobody"
serving with uuid 9e0d0953-e246-467b-b637-29ca8b1ad776
Debug logs will be stored in /opt/snare/snare.log
Error logs will be stored in /opt/snare/snare.err
Traceback (most recent call last):
  File "/usr/bin/snare", line 4, in <module>
    __import__('pkg_resources').run_script('Snare==0.3.0', 'snare')
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 657, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1444, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/lib/python3.6/site-packages/Snare-0.3.0-py3.6.egg/EGG-INFO/scripts/snare", line 204, in <module>
  File "/usr/lib/python3.6/site-packages/Snare-0.3.0-py3.6.egg/snare/server.py", line 91, in start
KeyError: '/status_404'
```

viskey98:
@t3chn0m4g3 Did you clone the page again after latest snare update?

t3chn0m4g3:
@viskey98 Yes. Rebuild the Docker container, used the latest setup instructions and cloned the page again.

viskey98:
@t3chn0m4g3 can you please show your clone output?

t3chn0m4g3:
cloner says it cannot resolve `example.com`, however the container is able to resolve it just fine:
```
/opt/snare # clone --target example.com

    ______ __      ______ _   ____________
   / ____// /     / __  // | / / ____/ __ \
  / /    / /     / / / //  |/ / __/ / /_/ /
 / /___ / /____ / /_/ // /|  / /___/ _, _/
/_____//______//_____//_/ |_/_____/_/ |_|
    
    
[...]

PR #155, opened 2018-08-02
fix 'status_404' with request containing prefixed protocol

viskey98:
@t3chn0m4g3 found out that when we clone pages like:
`clone --target http://example.com` and run snare.
We face `status_404` error in snare.
This pr will fix that issue #154


t3chn0m4g3:
@viskey98 Thank you!

PR #162, opened 2018-09-19
possibility to change default directory, fixes #134

### LINKED ISSUE #134: Add possibility to change default directory
Right now we always use `/opt/snare/` for storing all the snare files like config and uuid. It might be good to add the possibility to change this folder.

abhisharma404:
fixes #134 
This pull request adds the possibility to change the default directory. The default directory is currently set to `opt/snare` . At times user may want to change the default path, this PR gives them the possibility to change their default path by using a `--path <directory_path>` argument.

- Cloner can now take a `--path` argument, hence user can specify where they want to store their cloned webpage. If no `--path` argument is provided, cloner saves the cloned files in the default path, `opt/snare` .

- Snare config, uuid, pid is now saved into the path specified by the user else is stored in the default path, `/opt/snare` .

- While running Snare, user can specify their directory using `--path` argument .

- `add_meta_tag` in `test_add_meta_tag.py` is updated since `add_meta_tag` now takes 4 arguments instead of 3 (`default_path` is being passed) .

**Example :** 
1. `sudo clone --target http://example.com --path /root/file1` will save all the cloned web page in `root/file1/snare/pages/example.com`
2. snare config, pid and uid files will be saved in `root/file1/snare`
3. log path if specified will be saved in the specified path else will follow the path directory, here `root/file1/snare/clone.err` 
4. `sudo snare --page-dir example.com --path /root/file1` will retrieve the web pages from `root/file1/snare/pages/example.com`

Checklist : 

- [x] My branch is up-to-date with the Upstream master branch.
- [x] I have gone through the code base carefully before making this PR.

### DIFF
--- a/bin/clone
+++ b/bin/clone
@@ -38,15 +38,26 @@ def main():
     parser.add_argument(
         "--css-validate", help="set whether css validation is required", type=str_to_bool, default=None
     )
+    parser.add_argument("--path", help="path where the clone page will be saved", required=False, default=None)
     args = parser.parse_args()
-    if args.log_path:
-        log_err = args.log_path + "clone.err"
+    if args.path:
+        default_path = os.path.join(args.path, 'snare')
     else:
-        log_err = "/opt/snare/clone.err"
+        default_path = '/opt/snare'
+    if args.log_path:
+        log_err = os.path.join(args.log_path, 'clone.err')
+    else:
+        log_err = os.path.join(default_path, 'clone.err')
+
+    # Creates directory for log files
+    if not os.path.isfile(log_err):
+        if not os.path.exists(os.path.dirname(log_err)):
+            os.makedirs(os.path.dirname(log_err))
+
     logger.Logger.create_clone_logger(log_err, __package__)
     print("Error logs will be stored in {}\n".format(log_err))
     try:
-        cloner = Cloner(args.target, int(args.max_depth), args.css_validate)
+        cloner = Cloner(args.target, int(args.max_depth), args.css_validate, default_path)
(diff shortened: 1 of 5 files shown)


bin/clone
```diff
@@ -38,15 +38,26 @@ def main():
     parser.add_argument(
         "--css-validate", help="set whether css validation is required", type=str_to_bool, default=None
     )
+    parser.add_argument("--path", help="path where the clone page will be saved", required=False, default=None)
     args = parser.parse_args()
+    if args.path:
+        default_path = os.path.join(args.path, 'snare')
+    else:
+        default_path = '/opt/snare'
```
afeena: probably it's better to set it as default in options instead of None?

bin/snare
```diff
@@ -150,10 +149,15 @@ if __name__ == '__main__':
     parser.add_argument("--server-header", help="set server-header", default='nignx/1.3.8')
     parser.add_argument("--no-dorks", help="disable the use of dorks", type=str_to_bool,  default=True)
     parser.add_argument("--log-dir", help="path to directory of the log file", default='/opt/snare/')
+    parser.add_argument("--path", help="path to the directory of the cloned page", required=False, default=None)
     args = parser.parse_args()
-    base_path = '/opt/snare/'
-    base_page_path = '/opt/snare/pages/'
-    snare_uuid = snare_setup()
+    if not args.path:
+        base_path = '/opt/snare/'
```
afeena: same here, can't be set as default in parameters

rnehra01:
@abhisharma404 will you be able to make the suggested changes?

PR #174, opened 2018-11-27
Make dockerfile more flexible

leftshift:
A few improvements to the Dockerfile:
* Move page url to build arg `PAGE_URL`
* environment variable for tanner domain `TANNER`
* environment variable for snare listening port `PORT`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM alpine
 
-ARG PAGE_URL=example.org
+ARG PAGE_URL=example.com
+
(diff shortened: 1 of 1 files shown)


leftshift:
Ah, hang on, don't merge yet, there seems to be a problem with page names

leftshift:
This should do it. At build time, the build argument is also saved to an environment variable, which is available at runtime

glaslos:
Closing in favor of #177

PR #177, opened 2018-12-15
Improved docker

glaslos:
General cleanup of the docker file.
I dropped support for network interface in favor or removing a lot of dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,22 @@
-FROM alpine
+FROM python:3.6-alpine3.8
 
 ARG PAGE_URL=example.com
 
 ENV PAGE_URL $PAGE_URL
 ENV PORT 80
-ENV TANNER 127.0.0.1
-  
-# Setup apt
-RUN apk -U --no-cache add \
-               build-base \
-               git \
-               linux-headers \
-               python3 \
-               python3-dev && \
-# Setup Snare 
-    git clone --depth=1 https://github.com/mushorg/snare /opt/snare && \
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -1,28 +1,22 @@
...
-    rm -rf /opt/snare && \
-    clone --target http://example.com && \
-# Clean up
-    apk del --purge \
-            build-base \
-            linux-headers \
-            python3-dev && \
+FROM python:3.6-alpine3.8
+
+ARG PAGE_URL=example.com
```
mainnika: Isn't it better to put ARG and ENV commands just before «RUN clone» to keep build context as full as possible?

bin/snare
```diff
@@ -136,8 +135,7 @@ if __name__ == '__main__':
     page_group.add_argument("--list-pages", help="list available pages", action='store_true')
     parser.add_argument("--index-page", help="file name of the index page", default='index.html')
     parser.add_argument("--port", help="port to listen on", default='8080')
-    parser.add_argument("--interface", help="interface to bind to")
-    parser.add_argument("--host-ip", help="host ip to bind to", default='localhost')
+    parser.add_argument("--host-ip", help="host ip to bind to", default='127.0.0.1')
```
mainnika: Any problems with localhost?

glaslos: This was done to accommodate this change: https://github.com/mushorg/snare/pull/177/files#diff-a17cac3025400a430dedc6984959b596L195
Not sure if it's actually required.

PR #185, opened 2019-03-07
Changes in colner.py

sannanansari:
It was by default removing `\` in `cloner.py`.

### DIFF
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -33,10 +33,10 @@ class Cloner(object):
     def add_scheme(url):
         if yarl.URL(url).scheme:
             new_url = yarl.URL(url)
-            err_url = yarl.URL(url + '/status_404')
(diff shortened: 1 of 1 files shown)


sannanansari:
@rnehra01 is this correct? But however issue is created after changing `cloner.py`.

sannanansari:
@rnehra01 I have made changes in `cloner.py`. Is this correct?

rnehra01:
I'll review it tonight.

rnehra01:
What are these changes for?

sannanansari:
After going through error. I have found that extra `\` is causing an error. So after uderstanding `cloner.py` if found out it was by default adding extra `\`. So I remove it and it works fine.

rnehra01:
Firstly, I guess you mean `/` not `\`. And it is required in cases when URL does not have a trailing `/` like `example.com`.

PR #186, opened 2019-03-08
Both versions example.com and example.com/ are serving

amanjiofficial:


rnehra01:
I'll review it tonight.

amanjiofficial:
@rnehra01 okay

snare/cloner.py
```diff
@@ -31,12 +31,22 @@ def __init__(self, root, max_depth, css_validate):
 
     @staticmethod
     def add_scheme(url):
```
glaslos: can you add a test for this function?

snare/cloner.py
```diff
@@ -31,12 +31,22 @@ def __init__(self, root, max_depth, css_validate):
...
+        test_url = str(yarl.URL(url))
```
rnehra01: What is the use of this variable?

amanjiofficial: yarl URL cannot be accessed by index so 'test_url' has string version of the URL thus we can directly go to last index and check for presence of path.

amanjiofficial: Yes we should check conditions only for variable 'err_url' and we do not need to check it for 'new_url'.

amanjiofficial:
I have done the required changes. Please review

snare/cloner.py
```diff
@@ -31,11 +31,16 @@ def __init__(self, root, max_depth, css_validate):
...
     @staticmethod
     def add_scheme(url):
+        test_string = str(url)
         if yarl.URL(url).scheme:
             new_url = yarl.URL(url)
-            err_url = yarl.URL(url + '/status_404')
+            err_url = yarl.URL(url + 'status_404')
         else:
             new_url = yarl.URL('http://' + url)
+            err_url = yarl.URL('http://' + url + 'status_404')
```
rnehra01: It will produce wrong results. Move this condition at start and just remove the trailing `/`

snare/cloner.py
```diff
@@ -31,11 +31,16 @@ def __init__(self, root, max_depth, css_validate):
...
+        if test_string[-1] == '/':
```
rnehra01: It will produce wrong results. Move this condition at start and just remove the trailing `/`

amanjiofficial:
@rnehra01  Rather than removing the trailing ' _**/**_ '  I have added ' _**/**_ ' if it is not present. I have checked the code for possible cases. Please review for changes. Thank you

amanjiofficial:
@rnehra01 pls review the code

snare/cloner.py
```diff
@@ -31,12 +31,18 @@ def __init__(self, root, max_depth, css_validate):
 
     @staticmethod
     def add_scheme(url):
+        test_string = str(url)
```
rnehra01: `url` is already a string, right?

snare/cloner.py
```diff
@@ -31,12 +31,18 @@ def __init__(self, root, max_depth, css_validate):
...
+        if test_string[-1] == '/':
+            b_url = yarl.URL(url)
+        else:
+            b_url = yarl.URL(url + '/')
+        url = b_url.human_repr()
```
rnehra01: Keep the url as a string, no need of conversion to URL object and then just add the trailing space.

Issue #187, opened 2019-03-09
Create some basic test coverage


glaslos:
There is plenty of opportunity to add tests to SNARE. We make a lot of assumptions and with GSoC approaching, we should make sure the critical paths are covered.

amanjiofficial:
Can you please describe the test to be added with an example.

glaslos:
You can find examples of tests here:
https://github.com/mushorg/snare/tree/master/snare/tests
First step should be finding out where we lack coverage. Maybe use a
service like coverall?

amanjiofficial:
I am working on this issue I will update soon

PR #192, opened 2019-03-13
test tanner-parse-response

rjt-gupta:
Added test for `parse_tanner_response`.
Increased test coverage of `tanner_handler.py`  to 92%.
Refs - #187

### DIFF
--- a/snare/tests/test_parse_tanner_response.py
+++ b/snare/tests/test_parse_tanner_response.py
@@ -73,6 +73,8 @@ class TestParseTannerResponse(unittest.TestCase):
         expected_result = [self.expected_content, self.content_type, {}, 200]
         self.assertCountEqual(real_result, expected_result)
 
+    def test_parse_type_two_with_headers(self):
+
         self.detection = {
             "type": 2,
             "payload": {
--- a/snare/tests/test_parse_tanner_response.py
+++ b/snare/tests/test_parse_tanner_response.py
@@ -40,6 +40,7 @@ class TestParseTannerResponse(unittest.TestCase):
         self.detection = None
         self.expected_content = None
         self.call_content = None
+        self.expected_header = None
 
     def test_parse_type_one(self):
         self.detection = {"type": 1}
@@ -85,15 +86,17 @@ class TestParseTannerResponse(unittest.TestCase):
                 }
             }
         }
+        self.expected_content = b'test.png'
+        self.content_type = 'image/png'
+        self.expected_header = {"content-type": "multipart/form-data"}
 
         async def test():
             (self.res1, self.res2,
              self.res3, self.res4) = await self.handler.parse_tanner_response(self.requested_name, self.detection)
 
         self.loop.run_until_complete(test())
-        expected_header = {"content-type": "multipart/form-data"}
         real_result = [self.res1, self.res2, self.res3, self.res4]
-        expected_result = [b'test.png', 'image/png', expected_header, 200]
+        expected_result = [self.expected_content, self.content_type, self.expected_header, 200]
         self.assertCountEqual(real_result, expected_result)
 
     def test_parse_type_three(self):
--- a/snare/tests/test_parse_tanner_response.py
+++ b/snare/tests/test_parse_tanner_response.py
@@ -72,9 +72,7 @@ class TestParseTannerResponse(unittest.TestCase):
 
     def test_parse_type_one_error(self):
         self.detection = {"type": 1}
-        meta_content = {"/index.html": {}}
         self.requested_name = 'something/'
-        self.handler = TannerHandler(self.args, meta_content, self.uuid)
         self.expected_status_code = 404
 
         async def test():
@@ -135,14 +133,12 @@ class TestParseTannerResponse(unittest.TestCase):
         self.detection = {
             "type": 2,
             "payload": {
-                "page": "/index.html",
+                "page": "/something",
                 "value": "test"
             }
         }
-        meta_content = {"/index.html": {}}
-        self.handler = TannerHandler(self.args, meta_content, self.uuid)
         self.expected_content = b'<html><body><div>test</div></body></html>'
-        self.content_type = r'text\html'
+        self.content_type = r'text/html'
 
         async def test():
             (self.res1, self.res2,
@@ -184,6 +180,18 @@ class TestParseTannerResponse(unittest.TestCase):
 
         self.loop.run_until_complete(test())
         self.handler.html_handler.handle_content.assert_called_with(self.call_content)
+
+    def test_parse_exception(self):
+        self.detection = {}
+        self.call_content = b'<html><body></body></html>'
+        self.expected_content = self.page_content
+
+        async def test():
+            (self.res1, self.res2,
+             self.res3, self.res4) = await self.handler.parse_tanner_response(self.requested_name, self.detection)
+
+        with self.assertRaises(KeyError):
+            self.loop.run_until_complete(test())
 
     def tearDown(self):
         shutil.rmtree(self.main_page_path)
--- a/snare/tests/test_parse_tanner_response.py
+++ b/snare/tests/test_parse_tanner_response.py
@@ -73,6 +73,27 @@ class TestParseTannerResponse(unittest.TestCase):
         expected_result = [self.expected_content, self.content_type, {}, 200]
         self.assertCountEqual(real_result, expected_result)
 
+        self.detection = {
+            "type": 2,
+            "payload": {
+                "page": "",
+                "value": "test.png",
+                "headers": {
+                    "content-type": "multipart/form-data"
+                }
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


snare/tests/test_parse_tanner_response.py
```diff
@@ -73,6 +73,29 @@ def test_parse_type_two(self):
...
+        }
+
+        async def test():
+            (self.res1, self.res2,
+             self.res3, self.res4) = await self.handler.parse_tanner_response(self.requested_name, self.detection)
+
+        self.loop.run_until_complete(test())
+        expected_header = {"content-type": "multipart/form-data"}
+        real_result = [self.res1, self.res2, self.res3, self.res4]
+        expected_result = [b'test.png', 'image/png', expected_header, 200]
```
rnehra01: Use variables `self.expected_content`, `self.content_type` and assign after `detection`. Also I think to maintain consistency we can make `self.expected_header` and `self.expected_status_code` in setUp method.

rnehra01:
@rjt-gupta Can you try to add the remaining cases to make coverage 100%?

rjt-gupta:
Yes, Im working on it.

rjt-gupta:
The expected status code is `200` throughout the test. I don't think we should `self.expected_sataus_code` in setUp method. what do you think? @rnehra01

rnehra01:
Look [here](https://github.com/mushorg/snare/blob/master/snare/tanner_handler.py#L88).

rjt-gupta:
Oh yes, I missed this part as this part is not covered in the test yet. Will do the change surely :)

rjt-gupta:
Done.

snare/tests/test_parse_tanner_response.py
```diff
@@ -54,6 +56,36 @@ def test_parse_type_one(self):
...
+             self.res3, self.res4) = await self.handler.parse_tanner_response(self.requested_name, self.detection)
+
+        self.loop.run_until_complete(test())
+        real_result = [self.res1, self.res2, self.res3, self.res4]
+        expected_result = [self.page_content, self.content_type, {}, 200]
+        self.assertCountEqual(real_result, expected_result)
+
+    def test_parse_type_one_error(self):
+        self.detection = {"type": 1}
+        meta_content = {"/index.html": {}}
```
rnehra01: This can be removed.

snare/tests/test_parse_tanner_response.py
```diff
@@ -54,6 +56,36 @@ def test_parse_type_one(self):
...
+        self.requested_name = 'something/'
+        self.handler = TannerHandler(self.args, meta_content, self.uuid)
```
rnehra01: This can be removed too.

snare/tests/test_parse_tanner_response.py
```diff
@@ -105,17 +185,5 @@ def test_call_handle_html(self):
         self.loop.run_until_complete(test())
         self.handler.html_handler.handle_content.assert_called_with(self.call_content)
 
-    def test_parse_exception(self):
```
rnehra01: Why did you remove this?

snare/tests/test_parse_tanner_response.py
```diff
@@ -73,6 +105,54 @@ def test_parse_type_two(self):
...
+        real_result = [self.res1, self.res2, self.res3, self.res4]
+        expected_result = [self.expected_content, self.content_type, self.expected_header, 200]
+        self.assertCountEqual(real_result, expected_result)
+
+    def test_parse_type_two_error(self):
+
+        self.detection = {
+            "type": 2,
+            "payload": {
+                "page": "/index.html",
```
rnehra01: In this case also you can simulate the error using 
```suggestion
                "page": "/something",
```
and remove line [142](https://github.com/mushorg/snare/pull/192/files#diff-5a32a1b31bb3574ea82f22145d7864a7R142) and line 143

snare/tests/test_parse_tanner_response.py
```diff
@@ -73,6 +105,54 @@ def test_parse_type_two(self):
...
+                "value": "test"
+            }
+        }
+        meta_content = {"/index.html": {}}
+        self.handler = TannerHandler(self.args, meta_content, self.uuid)
+        self.expected_content = b'<html><body><div>test</div></body></html>'
+        self.content_type = r'text\html'
```
rnehra01: I think it should be 
```suggestion
        self.content_type = r'text/html'
```
Right?
And can you first make a separate PR to fix this https://github.com/mushorg/snare/blob/c04984110c15d8fbe031a77d51af7f17a423ca5b/snare/tanner_handler.py#L109

snare/tests/test_parse_tanner_response.py
rjt-gupta: The coverage is reached using other functions so need for this one.

snare/tests/test_parse_tanner_response.py
rjt-gupta: Roger!

snare/tests/test_parse_tanner_response.py
rjt-gupta: Yes, even though I have simulated the test-case using `something/` but there is no test that demonstrates this case which is also a possibility. 
But your decision would be final, should I remove this?

snare/tests/test_parse_tanner_response.py
rjt-gupta: PR submitted.

rjt-gupta:
Coverage is at 99%. How should I test this [line](https://github.com/mushorg/snare/blob/c04984110c15d8fbe031a77d51af7f17a423ca5b/snare/tanner_handler.py#L58).

snare/tests/test_parse_tanner_response.py
rnehra01: This case need not be handled here, it can be checked in `cloner.py`. So I think it's good to keep it simple here.

snare/tests/test_parse_tanner_response.py
rnehra01: I think this test is for [this](https://github.com/mushorg/snare/blob/master/snare/tanner_handler.py#L75) when detection is empty then raises a `keyError`. Since the code to handle is not written (which is because is unlikely to happen), that's why it's not present in the coverage. So I guess no harm in keeping this.

snare/tests/test_parse_tanner_response.py
rnehra01: Fix it.

rnehra01:
> Coverage is at 99%. How should I test this [line](https://github.com/mushorg/snare/blob/c04984110c15d8fbe031a77d51af7f17a423ca5b/snare/tanner_handler.py#L58)?

I guess you can make `error()` a mock method and use `assert_called`.

snare/tests/test_parse_tanner_response.py
rjt-gupta: Okay,

rjt-gupta:
Updated the changes.

snare/tests/test_submit_data.py
```diff
@@ -6,10 +6,14 @@
 import json
 import yarl
 import aiohttp
+import logging
+from unittest import mock
 from snare.utils.asyncmock import AsyncMock
 from snare.tanner_handler import TannerHandler
 from snare.utils.page_path_generator import generate_unique_path
 
+logger = logging.getLogger(__name__)
```
rnehra01: Can you place it in `setUp`?

snare/tests/test_submit_data.py
```diff
@@ -65,6 +69,21 @@ def test_event_result(self):
         self.loop.run_until_complete(test())
         self.assertEqual(self.result, dict(detection={'type': 1}, sess_uuid="test_uuid"))
 
+    def test_submit_data_error(self):
+
+        async def mock_json_decode_error(data):
+            logger.error('Error submitting data: JSONDecodeError {}'.format(data))
+
+        self.handler = mock.Mock()
+        self.handler.submit_data = mock_json_decode_error
```
rnehra01: I think you should use [this](https://github.com/mushorg/snare/blob/master/snare/tests/test_submit_data.py#L69) instead of mocking the entire `submit_data` method.

snare/tests/test_submit_data.py
```diff
@@ -65,6 +69,21 @@ def test_event_result(self):
         self.loop.run_until_complete(test())
         self.assertEqual(self.result, dict(detection={'type': 1}, sess_uuid="test_uuid"))
 
+    def test_submit_data_error(self):
+
+        async def mock_json_decode_error(data):
+            logger.error('Error submitting data: JSONDecodeError {}'.format(data))
```
rnehra01: Usually, logger creates a file for errors, is this also doing the same? If so please do the cleanup.

rjt-gupta:
Done. 100% coverage reached.

Issue #182, opened 2019-03-03
GSoC 2019 - Getting Started


rnehra01:
To get started

- Clone the repo and try to run the tool
- Understand the code through documentation
- The cloner part of snare is NOT perfect, try to clone different website and see if some part of the webpage(images, css, js, realtive path handling) is not properly cloned. (e.g I found #183 )
- Try to complete missing documentation
- Try to add missing tests for methods (E.g https://github.com/mushorg/snare/pull/186#discussion_r264003865)

afeena:
Outdated :)

PR #194, opened 2019-03-14
clear query in url while cloning

shin2903:
## Description
Fix for #182.
Cloner preserve query string in url before creating `filename`, while snare remove query string before serving content. This inconsistency leads to 404 response.
## How to reproduce
1. `clone --target http://readms.net --max-depth 1` 
2. `sudo snare --port 8080 --page-dir readms.net`
3. Visit http://localhost:8080/
## Solution
Remove query string in url before creating `filename`

### DIFF
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -102,6 +102,8 @@ class Cloner(object):
 
     def _make_filename(self, url):
         host = url.host
+        # clear query
+        url = url.with_query(None)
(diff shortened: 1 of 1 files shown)


rnehra01:
But certain URL server different content based on query parameters. So I think it needs some thinking to solve this. Please create an issue for it.

rnehra01:
#199

Issue #196, opened 2019-03-15
Handle query string in URL while cloning


shin2903:
Currently, snare would return `404 Not found` when serving resources with query string in url .

rnehra01:
@shin2903 Can you provide the steps to regenerate the issue?

shin2903:
## How to reproduce
1. `clone --target https://readms.net --max-depth 1` 
2. `sudo snare --port 8080 --page-dir readms.net`
3. Visit http://localhost:8080/, we see that the cloned page is not rendered correctly

rnehra01:
#199

PR #199, opened 2019-03-25
check url with query when detection = 1

shin2903:
Fix for #196. 
Check url with query when `detection = 1`. This allows serving cloned content with query in url.

### DIFF
--- a/snare/tanner_handler.py
+++ b/snare/tanner_handler.py
@@ -78,7 +78,7 @@ class TannerHandler():
             if query_start != -1:
                 possible_requests.append(requested_name[:query_start])
 
-            found = False
+            file_name = None
             for requested_name in possible_requests:
                 if requested_name == '/':
                     requested_name = self.run_args.index_page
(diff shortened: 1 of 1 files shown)


snare/tanner_handler.py
```diff
@@ -73,18 +73,27 @@ def create_data(self, request, response_status):
...
+            possible_requests = [requested_name]
             query_start = requested_name.find('?')
             if query_start != -1:
-                requested_name = requested_name[:query_start]
-            if requested_name == '/':
-                requested_name = self.run_args.index_page
-            try:
+                possible_requests.append(requested_name[:query_start])
+
+            found = False
```
rnehra01: Instead of `found` variable you can use `file_name = None`

rnehra01:
The logic looks fine but to merge it, I need to do more tests.

snare/tanner_handler.py
shin2903: Yes I have updated the code.

PR #204, opened 2019-04-01
add server header in middleware

shin2903:
Fix for #7 
## Steps to reproduce
1. `sudo clone --target http://example.com`
2. `sudo snare --port 8080 --page-dir example.com`
3. `curl --head http://localhost:8080/notfound`. Response:
```
HTTP/1.1 200 OK
Content-Type: text/html; charset=utf-8
Content-Length: 1237
Date: Mon, 01 Apr 2019 16:49:00 GMT
Server: Python/3.6 aiohttp/3.4.4
```
Server header above is not changed to default `nginx`

### DIFF
--- a/snare/middlewares.py
+++ b/snare/middlewares.py
@@ -20,10 +20,12 @@ class SnareMiddleware():
         async def error_middleware(request, handler):
             try:
                 response = await handler(request)
-                override = overrides.get(response.status)
+                status = response.status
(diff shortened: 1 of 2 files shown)


rnehra01:
The response code is `200` while it should be `404`. Can you raise an issue for that and look into that before merging this?

shin2903:
Sure

PR #218, opened 2019-08-05
Improved Logging

rjt-gupta:


snare/cloner.py
```diff
@@ -127,6 +128,7 @@ def _make_filename(self, url):
             self.visited_urls.append(current_url.human_repr())
             file_name, hash_name = self._make_filename(current_url)
             print('name: ', file_name)
```
afeena: Since you want to move everything into logs, maybe it's better to remove print?

snare/cloner.py
```diff
@@ -127,6 +128,7 @@ def _make_filename(self, url):
...
+            self.logger.debug('Cloned file: %s', file_name)
```
afeena: Only error log exists for cloner, you have to set up logger with another level to have debug messages

snare/cloner.py
```diff
@@ -150,6 +152,7 @@ def _make_filename(self, url):
                     index_fh.write(data)
                 if content_type == 'text/css':
                     css = cssutils.parseString(data, validate=self.css_validate)
+                    self.logger.info('Validating the CSS..')
```
afeena: This info is useless and provides multiple entries in the log.

snare/cloner.py
rjt-gupta: Yup.

snare/cloner.py
rjt-gupta: Done

rjt-gupta:
@afeena Anything else needs to be done?

snare/cloner.py
```diff
@@ -14,6 +14,8 @@
 
 class Cloner(object):
     def __init__(self, root, max_depth, css_validate):
+        self.logger = logging.getLogger(__name__)
+        self.logger.setLevel(logging.DEBUG)
```
afeena: This doesn't work (at least for me). Have you run it?

The handler of the parent logger, which created in cloner bin file has a level ERROR, guess this is the problem

afeena: This doesn't work (at least for me). Have you checked this?

I guess the problem might be in the parent logger, which has a level ERROR

rjt-gupta: Umm. It works for me though.

Let me write a test for this.

afeena: I have no idea why it doesn't work for me... Let me test on another env

rjt-gupta: Alright :)

afeena: Ok, now there is no print in console, but still, there are no cloned page logs in log

rjt-gupta: I am not sure what's wrong. Are tests passing on your side?

afeena: I run `clone --target example.com` and I want to see 2 file names in the log. And the log file is empty

afeena: Tests are passing

afeena: When you run cloner, do you have records in the file?

rjt-gupta: Sorry, I was not able to look into it because of my exams.

Will try to look today :)

afeena: If you now set the level in the logger creation, you don't need it here

snare/utils/logger.py
```diff
@@ -48,6 +48,6 @@ def create_clone_logger(err_filename, logger_name):
         )
         # ERROR log to 'clone.err'
         error_log_handler = logging.handlers.RotatingFileHandler(err_filename, encoding='utf-8')
-        error_log_handler.setLevel(logging.ERROR)
+        error_log_handler.setLevel(logging.DEBUG)
```
afeena: please change the variable names and file name, because now it's ambiguous, it's not only errors anymore

rjt-gupta: Done.

snare/cloner.py
rjt-gupta: I tried removing it but logger only works when this and parent logger are both set to DEBUG.

rjt-gupta: @afeena Can you have a look?

afeena: @rjt-gupta only after 26th of September, until then I am not available. If someone can take a look before, would be nice :) @rnehra01 @glaslos

snare/utils/logger.py
```diff
@@ -41,13 +41,13 @@ def create_logger(debug_filename, err_filename, logger_name):
...
         logger = logging.getLogger(logger_name)
         formatter = logging.Formatter(
             fmt='%(asctime)s %(levelname)s:%(name)s:%(funcName)s: %(message)s', datefmt='%Y-%m-%d %H:%M:%S'
         )
-        # ERROR log to 'clone.err'
-        error_log_handler = logging.handlers.RotatingFileHandler(err_filename, encoding='utf-8')
-        error_log_handler.setLevel(logging.ERROR)
-        error_log_handler.setFormatter(formatter)
-        logger.addHandler(error_log_handler)
+        # logs to 'clone.err'
```
afeena: the comment doesn't match the filename :)

PR #235, opened 2020-02-19
Fixed documentation quick start section

puneet29:
Fixed documentation. Pip installations must be done to the global pip environment with superuser privileges, to omit the error.

### DIFF
--- a/docs/quick-start.rst
+++ b/docs/quick-start.rst
@@ -28,7 +28,7 @@ Getting started
 
 1. Get SNARE: ``git clone https://github.com/mushorg/snare.git`` and ``cd snare``
 
-2. Install requirements: ``pip3 install -r requirements.txt``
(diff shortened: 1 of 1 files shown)


docs/quick-start.rst
```diff
@@ -28,7 +28,7 @@ Getting started
 
 1. Get SNARE: ``git clone https://github.com/mushorg/snare.git`` and ``cd snare``
 
-2. Install requirements: ``pip3 install -r requirements.txt``
+2. Install requirements: ``sudo pip3 install -r requirements.txt``
```
afeena: Do we really want to install libraries with sudo?

puneet29: Hi @afeena 
Yes, actually the libraries are to be installed in the global pip environment because while running the subsequent commands it might result in an error in some cases.
You can verify this from the README.md as well. It has also been changed accordingly. Also check #178

PR #239, opened 2020-02-25
Allow specifying a cloner target with path/qs

adepasquale:
Cloner can now use a starting page requiring both a path and a query string


afeena:
Thanks for you pr :)
Could you please check the broken test? Seems like it doesn't work with relative links. 
Maybe you can use something like https://yarl.readthedocs.io/en/latest/api.html#yarl.URL.build to create URL with a scheme?

PR #241, opened 2020-02-25
Clone and serve headers

adepasquale:
Some websites might have interesting headers. This pull request adds code to store such headers and subsequently serve them.

The default `Server: nginx/1.3.8` header is also removed in favour of the "cloned" one, but it's still possible to specify a different server with the command line switch.

### DIFF
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -43,11 +43,16 @@ class Cloner(object):
 
     @staticmethod
     def get_headers(response):
-        # TODO probably need more...
         ignored_headers_lowercase = [
-            "connection",
-            "content-length",
-            "content-type",
+            'age',
+            'cache-control',
+            'connection',
+            'content-encoding',
+            'content-length',
+            'date',
+            'etag',
+            'expires',
+            'x-cache',
         ]
 
         headers = []
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -157,22 +157,21 @@ class Cloner(object):
                 # TODO might not be needed anymore if we can rely on the header
                 content_type = response.content_type
                 data = await response.read()
-
             except (aiohttp.ClientError, asyncio.TimeoutError) as client_error:
                 self.logger.error(client_error)
             else:
                 await response.release()
+
             if data is not None:
                 self.meta[file_name]['hash'] = hash_name
                 self.meta[file_name]['headers'] = headers
                 self.meta[file_name]['content_type'] = content_type
                 self.counter = self.counter + 1
+
                 if content_type == 'text/html':
                     soup = await self.replace_links(data, level)
                     data = str(soup).encode()
-                with open(os.path.join(self.target_path, hash_name), 'wb') as index_fh:
-                    index_fh.write(data)
-                if content_type == 'text/css':
+                elif content_type == 'text/css':
                     css = cssutils.parseString(data, validate=self.css_validate)
                     for carved_url in cssutils.getUrls(css):
                         if carved_url.startswith('data'):
@@ -182,6 +181,9 @@ class Cloner(object):
                             carved_url = self.root.join(carved_url)
                         if carved_url.human_repr() not in self.visited_urls:
                             await self.new_urls.put((carved_url, level + 1))
+
+                with open(os.path.join(self.target_path, hash_name), 'wb') as index_fh:
+                    index_fh.write(data)
 
     async def get_root_host(self):
         try:
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -41,6 +41,21 @@ class Cloner(object):
         err_url = new_url.with_path('/status_404').with_query(None).with_fragment(None)
         return new_url, err_url
 
+    @staticmethod
+    def get_headers(response):
+        # TODO probably need more...
+        ignored_headers_lowercase = [
+            "connection",
+            "content-length",
+            "content-type",
+        ]
+
+        headers = []
+        for key, value in response.headers.items():
+            if key.lower() not in ignored_headers_lowercase:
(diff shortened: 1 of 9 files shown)


snare/cloner.py
```diff
@@ -134,6 +149,8 @@ def _make_filename(self, url):
             content_type = None
             try:
                 response = await session.get(current_url, headers={'Accept': 'text/html'}, timeout=10.0)
+                headers = self.get_headers(response)
+                # TODO might not be needed anymore if we can rely on the header
                 content_type = response.content_type
```
adepasquale: If this prq makes sense, I can change the code to rely on the content-type found in the header instead of this one.

adepasquale:
Waiting for feedback before fixing tests.

snare/tanner_handler.py
```diff
@@ -127,7 +136,8 @@ def create_data(self, request, response_status):
                 content = payload_content['value'].encode('utf-8')
 
             if 'headers' in payload_content:
-                headers = payload_content['headers']
+                headers.update(payload_content['headers'])  # overwrite local headers with the tanner-provided ones
+                #headers.extend(payload_content['headers'])  # keep both headers with same name (undesired behavior)
```
glaslos: agreed, feel free to remove

adepasquale:
Fixed tests, also resolved TODO/FIXMEs except one that would require some Tanner-side changes and possibly introduce backward compatibility issues.

snare/server.py
```diff
@@ -89,8 +77,9 @@ def __init__(self, meta, run_args, snare_uuid, debug=False, keep_alive=75, **kwa
             app, loader=jinja2.FileSystemLoader(self.dir)
         )
         middleware = SnareMiddleware(
-            self.meta['/status_404']['hash'],
-            self.run_args.server_header
+            headers=self.meta['/status_404']['headers'],
```
afeena: I got KeyError with old-cloned page, works perfectly after re-cloning, maybe it would be good to handle this problem? At least show some meaningful error and mention that re-cloning is needed

afeena:
@adepasquale  Awesome! Looks and works well :) Just one minor comment (see above)

afeena:
@adepasquale just encountered a bug, I haven't tested it agains all attack, but such request `index.html?foo=/etc/passwd` cause the error:

```2020-02-28 14:29:28 ERROR:aiohttp.server:log_exception: Unhandled exception
Traceback (most recent call last):
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_protocol.py", line 410, in start
    await resp.prepare(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_response.py", line 300, in prepare
    return await self._start(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_response.py", line 608, in _start
    return await super()._start(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_response.py", line 367, in _start
    await writer.write_headers(status_line, headers)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/http_writer.py", line 109, in write_headers
    buf = _serialize_headers(status_line, headers)
  File "aiohttp/_http_writer.pyx", line 138, in aiohttp._http_writer._serialize_headers
  File "aiohttp/_http_writer.pyx", line 110, in aiohttp._http_writer.to_str
TypeError: Cannot serialize non-str key None```

in `master` such request work as usual. Could you please check?

File: snare/server.py (hunk 3 above)
Comment (adepasquale) [reply snare/server.py]: I made the rest of the new code backward-compatible but I missed this part. I will change it so that if the headers are not available they will be ignored.

**Comment by adepasquale** [comment]
@afeena I'm not able to reproduce the above serialization issue. Could you please post here (or dm/email) the full request/response and the website you used for cloning? Thanks!

**Comment by afeena** [comment]
@adepasquale 

I still have the same problem, steps to reproduce:

1. clone snare 
2. git checkout -b adepasquale-cloner_meta_headers master
3. git pull https://github.com/adepasquale/snare.git cloner_meta_headers
4. python3 setup.py install
5. clone --target example.com
6. snare --page-dir example.com --host-ip 0.0.0.0 --tanner 127.0.0.1 --auto-update False
7.`index.html` --> works
8.`index.html?foo=/etc/passwd` --> fails

`snare.err` logs:

```
2020-03-02 22:31:11 ERROR:aiohttp.server:log_exception: Unhandled exception
Traceback (most recent call last):
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_protocol.py", line 410, in start
    await resp.prepare(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_response.py", line 300, in prepare
    return await self._start(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_response.py", line 608, in _start
    return await super()._start(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_response.py", line 367, in _start
    await writer.write_headers(status_line, headers)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/http_writer.py", line 109, in write_headers
    buf = _serialize_headers(status_line, headers)
  File "aiohttp/_http_writer.pyx", line 138, in aiohttp._http_writer._serialize_headers
  File "aiohttp/_http_writer.pyx", line 110, in aiohttp._http_writer.to_str
TypeError: Cannot serialize non-str key None
```

I put `print(headers)` before this line 
https://github.com/mushorg/snare/pull/241/files#diff-9fa6d284aff3d2e6f8e4808d8681dd52R149

Result:
`<CIMultiDict('Content-Type': None)>`
I guess this might be the problem.


Thanks!

PR #238, opened 2020-02-24
Remove version restriction to prevent gitdb error

mzfr:
When we try to do `sudo docker-compose up` the container resulted in the following error:

```
Creating snare ... done
Attaching to snare
snare    | Traceback (most recent call last):
snare    |   File "/usr/local/bin/snare", line 4, in <module>
snare    |     __import__('pkg_resources').run_script('Snare==0.3.0', 'snare')
snare    |   File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 667, in run_script
snare    |     self.require(requires)[0].run_script(script_name, ns)
snare    |   File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1470, in run_script
snare    |     exec(script_code, namespace, namespace)
snare    |   File "/usr/local/lib/python3.6/site-packages/Snare-0.3.0-py3.6.egg/EGG-INFO/scripts/snare", line 29, in <module>
snare    |   File "/usr/local/lib/python3.6/site-packages/git/__init__.py", line 38, in <module>
snare    |     from git.exc import *                       # @NoMove @IgnorePep8
snare    |   File "/usr/local/lib/python3.6/site-packages/git/exc.py", line 9, in <module>
snare    |     from git.compat import UnicodeMixin, safe_decode, string_types
snare    |   File "/usr/local/lib/python3.6/site-packages/git/compat.py", line 16, in <module>
snare    |     from gitdb.utils.compat import (
snare    | ModuleNotFoundError: No module named 'gitdb.utils.compat'
```
But if we remove that version restriction the `docker-compose up`  command runs succesfully.

PR #242, opened 2020-02-25
Corrected the Python version in README

Parth1811:
Updated the python cersion in README.md [Refs](https://github.com/mushorg/tanner/issues/354#issuecomment-590316566)
Changed the version number in docker-compose.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Basic Concepts
 Getting started
 ---------------
 
-- You need Python3. We tested primarily with >=3.5
+- You need Python3.6 to run SNARE
 - This was tested with a recent Ubuntu based Linux.
 
 #### Steps to setup


requirements.txt
```diff
@@ -2,5 +2,5 @@ aiohttp==3.4.4
 aiohttp_jinja2==1.1.0
 beautifulsoup4==4.6.3
 cssutils==1.0.2
-gitpython==2.1.11
+gitpython
```
glaslos: Why are you unpinning the version?

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '2.3'
+version: '2.0'
```
glaslos: why downgrading the docker compose version?

README.md
```diff
@@ -23,7 +23,7 @@ Basic Concepts
 Getting started
 ---------------
 
-- You need Python3. We tested primarily with >=3.5
+- You need Python3.6.x or more to run SNARE
```
glaslos:
```suggestion
- You need Python 3.6 to run SNARE
```

requirements.txt
Parth1811: version 2.1.11 causes run time error in setup.py while using python 3.6 or higher. I unpinned the version to get the latest update of gitpython.

docker-compose.yml
Parth1811: Running `docker-compse build` gives the following error:-
```
parth@parth:(master)~/snare$ docker-compose build 
ERROR: Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

hence I followed the suugestion

requirements.txt
afeena: fixed in other PR #238 , please rebase you branch

Parth1811: @afeena Cool, I was trying to fix the exact same problem as in #238

docker-compose.yml
afeena: @Parth1811 which version of Docker are you using?
Please check this
https://docs.docker.com/compose/compose-file/

I can't get an error you are reporting, so probably you have old docker version?

Parth1811: @afeena  yes you are right, I was on an older version.

afeena:
Good! Thanks

Issue #224, opened 2019-12-16
Traceback on running snare


Aniket21mathur:
Got internal server error on visiting `localhost:8080`.

`snare.err`

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/aiohttp/web_protocol.py", line 390, in start
    resp = await self._request_handler(request)
  File "/usr/local/lib/python3.6/dist-packages/aiohttp/web_app.py", line 366, in _handle
    resp = await handler(request)
  File "/usr/local/lib/python3.6/dist-packages/aiohttp/web_middlewares.py", line 106, in impl
    return await handler(request)
  File "/usr/local/lib/python3.6/dist-packages/Snare-0.3.0-py3.6.egg/snare/middlewares.py", line 22, in error_middleware
    response = await handler(request)
  File "/usr/local/lib/python3.6/dist-packages/Snare-0.3.0-py3.6.egg/snare/server.py", line 50, in handle_request
    event_result = await self.tanner_handler.submit_data(data)
  File "/usr/local/lib/python3.6/dist-packages/Snare-0.3.0-py3.6.egg/snare/tanner_handler.py", line 63, in submit_data
    raise e
  File "/usr/local/lib/python3.6/dist-packages/Snare-0.3.0-py3.6.egg/snare/tanner_handler.py", line 56, in submit_data
    event_result = await r.json()
  File "/usr/local/lib/python3.6/dist-packages/aiohttp/client_reqrep.py", line 938, in json
    headers=self.headers)
aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: text/plain; charset=utf-8'
```
Possibly because the `content-type` is not passed to the json function or we can instead use `json.loads`.

@afeena @rjt-gupta @rnehra01 please have a look. Thanks!

afeena:
@Aniket21mathur I fixed the problem with default tanner server, you can check it now :)

Aniket21mathur:
@afeena thanks I tried to run snare again with the default tanner server `example.com` but it doesn't run. It stops as shown below in the image. Nothing in the `logs.err`, I am afraid I am not able to figure out why this is happening. 

Please guide. Thanks!
![Screenshot from 2019-12-18 09-53-41](https://user-images.githubusercontent.com/35092582/71055777-e7646880-217c-11ea-9393-87b852344434.png)

afeena:
[...]

Issue #201, opened 2019-03-30
KeyError 'hash' / 'Value' whle running `snare --port 8080 --page-dir "some site name"


ADI10HERO:
Hello i sucessully set up snare (example.com) worked but then cloning twitter,gmail and then my college website gave errors.
Twitter and Gmail were not clonned to begin with but the college website (http) was cloned but when i tried to run the above mentioned command it gave the following error : 

""" 
Traceback (most recent call last):
  File "/usr/local/bin/snare", line 4, in <module>
    __import__('pkg_resources').run_script('Snare==0.3.0', 'snare')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 658, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1445, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python3.6/dist-packages/Snare-0.3.0-py3.6.egg/EGG-INFO/scripts/snare", line 200, in <module>
  File "/usr/lib/python3.6/asyncio/base_events.py", line 473, in run_until_complete
    return future.result()
  File "/usr/local/lib/python3.6/dist-packages/Snare-0.3.0-py3.6.egg/snare/server.py", line 91, in start
KeyError: 'hash'
you are running the latest version
""""
Help wanted :)

rnehra01:
Can you provide the link your college website?

ADI10HERO:
Yes sure, well my college official website is bad (sorry) This is another website giving same error i think this will do? 
http://www.kjscecodecell.com/

rnehra01:
It is working fine for me. Can you show the commands you run?

erwanlr:
I have the same issue. Cloned a local site with `sudo clone --target http://wp.lab/`

Then ran `sudo snare --port 8080 --page-dir wp.lab` which resulted in:


```
Traceback (most recent call last):
  File "/usr/local/bin/snare", line 4, in <module>
    __import__('pkg_resources').run_script('Snare==0.3.0', 'snare')
  File "/Users/xxxx/Library/Python/3.7/lib/python/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/Users/xxxx/Library/Python/3.7/lib/python/site-packages/pkg_resources/__init__.py", line 1453, in run_script
    exec(script_code, namespace, namespace)
[...]

PR #243, opened 2020-02-25
Fix internal server error due to unexpected mimetype

mzfr:
I've fixed the `Internal Server Error` problem.

Issue #224 and my [comment](https://github.com/mushorg/snare/issues/224#issuecomment-590468689) on that explains the problem.

Also if setting `content_type=None` isn't considered good practice then 
we can set that to mimetype `text/plain` as reported in the traceback

@afeena @adepasquale please review this PR and let me know if any changers are required.

### DIFF
--- a/snare/tanner_handler.py
+++ b/snare/tanner_handler.py
@@ -53,7 +53,7 @@ class TannerHandler():
                     timeout=10.0
                 )
                 try:
-                    event_result = await r.json()
+                    event_result = await r.json(content_type=None)
                 except json.decoder.JSONDecodeError as e:
                     self.logger.error('Error submitting data: {} {}'.format(e, data))
                 finally:


adepasquale:
There should be no need to merge this. The issue is related to you using the default tanner instance at `tanner.mushmush.org`, which is failing for some unknown reasons. Please use your own tanner instance.

mzfr:
oh so running snare in one docker and running tanner in another would fix the issue?

mzfr:
Also looks like (others](https://github.com/mushorg/snare/issues/201#issuecomment-589694183) are also facing this issue.

afeena:
Maybe we could catch the tanner problem via this error and show some page (index page)? Which behaviour should we expect?
What do you think? Any ideas?

PR #246, opened 2020-02-27
Fixes #134

### LINKED ISSUE #134: Add possibility to change default directory

puneet29:


puneet29:
@afeena Can you please review the changes that I have made. I am working on making the tests work.

afeena:
duplicates #226

puneet29:
Its been 7 days since the original PR has not been merged or worked upon. @afeena

afeena:
@puneet29  @Aniket21mathur closed the original, so I reopen this one

afeena:
@puneet29 please fix conflicts first

bin/snare
```diff
@@ -69,18 +66,6 @@ def snare_setup():
         return snare_uuid
 
 
-def drop_privileges():
```
afeena: You shouldn't remove this function

puneet29: Do the new changes resolve this?

bin/clone
```diff
@@ -38,29 +36,36 @@ def main():
     parser.add_argument(
         "--css-validate", help="set whether css validation is required", type=str_to_bool, default=None
     )
+    parser.add_argument("--path", help="path to save the page to be cloned", required=False, default="/opt/")
```
afeena: If you keep /opt/ as default, you should use `sudo`. And I prefer to keep it as default. So Virtual env is optional, but not forced

puneet29:
Hey @afeena, there's another issue which bugs the developers while changing `bin/snare`.
When we fix a bug or change the code and test our code by running the snare command it pulls the latest copy from the server, removing all the bug fixes. I will raise a ticket for the same.

puneet29:
@afeena Is the README quickstart understandable or should it be changed?

bin/snare
```diff
@@ -148,14 +146,15 @@ if __name__ == '__main__':
     parser.add_argument("--server-header", help="set server-header", default='nginx/1.3.8')
     parser.add_argument("--no-dorks", help="disable the use of dorks", type=str_to_bool,  default=True)
     parser.add_argument("--log-dir", help="path to directory of the log file", default='/opt/snare/')
+    parser.add_argument("--path", help="path to save the page to be cloned", required=False, default='/opt/')
```
afeena: I suggest you reuse this path for both logs and snare files, what do you think?

bin/snare
```diff
@@ -148,14 +146,15 @@ if __name__ == '__main__':
...
     args = parser.parse_args()
-    base_path = '/opt/snare/'
-    base_page_path = '/opt/snare/pages/'
-    snare_uuid = snare_setup()
+    base_path = os.path.join(args.path,'snare')
+    base_page_path = os.path.join(base_path,'pages')
```
afeena: What if I want to use different paths fo pages and snare? I wanted to reuse cloned pages from /opt/ and I was unable to do that. Do we want this functionality?

bin/clone
```diff
@@ -38,29 +32,46 @@ def main():
...
+    default_path = os.path.join(args.path, 'snare')
+
     if args.log_path:
-        log_file = args.log_path + "clone.log"
+        log_file = os.path.join(args.log_path, "clone.log")
     else:
-        log_file = "/opt/snare/clone.log"
+        log_file = os.path.join(default_path, 'clone.log')
+
+    if(not os.path.realpath(default_path).startswith(os.path.expanduser('~')) or
```
afeena: to test permissions, you can also easily try to write something into the base path and catch the exception.
A user might have permissions not only to his own folder

afeena:
@puneet29  I have to think about it :)

bin/snare
puneet29: I'd same thing on my mind

bin/snare
puneet29: I have added the functionality that you're requesting

bin/clone
```diff
@@ -38,29 +32,55 @@ def main():
...
+
     if args.log_path:
-        log_file = args.log_path + "clone.log"
+        log_file = os.path.join(args.log_path, "clone.log")
     else:
-        log_file = "/opt/snare/clone.log"
+        log_file = os.path.join(default_path, 'clone.log')
+
+    try:
+        if not os.path.exists(default_path):
```
afeena: I'd suggest you make a separate function for checking permissions, to avoid code duplications, since you are using it multiple times

afeena:
I tested with next command

`snare --page-dir /opt/snare/pages/example.com --host-ip 0.0.0.0 --auto-update False --tanner 127.0.0.1 --path /home/afeena/snare` 
Snare runs normally, but I do not see a page when I open a browser, just a blank screen. And there is no errors in logs. COuld you please check?

bin/snare
```diff
@@ -147,15 +152,20 @@ if __name__ == '__main__':
     parser.add_argument("--update-timeout", help="update snare every timeout ", default='24H')
     parser.add_argument("--server-header", help="set server-header", default='nginx/1.3.8')
     parser.add_argument("--no-dorks", help="disable the use of dorks", type=str_to_bool,  default=True)
-    parser.add_argument("--log-dir", help="path to directory of the log file", default='/opt/snare/')
+    parser.add_argument("--path", help="path to save the page to be cloned", required=False, default='/opt/')
+    parser.add_argument("--page-path", help="Use only if the path not same as that of snare. Path to cloned pages",
```
afeena: maybe instead of adding new CLI argument (we have so many now and it's easy to be confused) it's better to allow use an absolute path for page-dir? What do you think? Smth like if page-dir is relative, merge it with base_path, otherwise use absolute

bin/snare
```diff
@@ -147,15 +152,20 @@ if __name__ == '__main__':
...
+                        required=False, default='none')
```
afeena: why using string `none` instead of `None`?

afeena:
Works good :) Just some minor comments on code style :)  I suggest you put everything regarding virtualenv into the snare docs and put the link in the README :)

bin/snare
```diff
@@ -147,15 +152,20 @@ if __name__ == '__main__':
     parser.add_argument("--update-timeout", help="update snare every timeout ", default='24H')
     parser.add_argument("--server-header", help="set server-header", default='nginx/1.3.8')
     parser.add_argument("--no-dorks", help="disable the use of dorks", type=str_to_bool,  default=True)
-    parser.add_argument("--log-dir", help="path to directory of the log file", default='/opt/snare/')
+    parser.add_argument("--path", help="path to save the page to be cloned", required=False, default='/opt/')
```
afeena: For me now this parameter a little bit misleading, when I set directory by myself, I set the final directory I want to use and snare creates additional directories. What do you think about it? Should we set the resulting directory or parent directory for the pages?

bin/clone
```diff
@@ -38,29 +32,55 @@ def main():
...
+            tempfile.write('')
+        os.remove(os.path.join(default_path, 'temp'))
+
+        if not os.path.exists(os.path.dirname(log_file)):
+            os.makedirs(os.path.dirname(log_file))
+        with open(os.path.join(os.path.dirname(log_file), 'temp'), 'w') as tempfile:
+            tempfile.write('')
+        os.remove(os.path.join(os.path.dirname(log_file), 'temp'))
+    except IOError as e:
+        if os.getuid() != 0:
```
afeena: this check is now redundant because you have already tested your permissions for the folders

bin/snare
puneet29: We should probably change this argument''s function to tell the base directory for snare. What do you think?

bin/snare
puneet29: Yeah, that seems only logical.

afeena:
@puneet29 Looks good! Could you please resolve conflict and I will merge?

puneet29:
Hey @afeena, Thank you for sticking around and guiding me. I've done the requested changes.

bin/snare
```diff
@@ -226,7 +230,8 @@ if __name__ == '__main__':
     loop = asyncio.get_event_loop()
     try:
         loop.run_until_complete(app.start())
-        drop_privileges()
+        if os.getuid() == 0:
```
afeena: Sorry, I missed this part. I would prefer to drop privileges for all users, even without root permissions

puneet29: Hey @afeena 

Here's what the output is when I do it for a normal user:
```
>>> os.setgid(wanted_group.gr_gid)
PermissionError: [Errno 1] Operation not permitted

Can you please help me out here. I am new to using such a technique. Does a normal run (not superuser) require dropping privileges? Thanks and regards.

afeena: My bad! Just keep you variant :)

puneet29: No biggie! Thanks a lot. :)

PR #247, opened 2020-02-28
catch and log unexpected error causing internal server error

mzfr:
@afeena As discussed for now I've added the exception so at least this error would be logged.

### DIFF
--- a/snare/server.py
+++ b/snare/server.py
@@ -49,9 +49,6 @@ class HttpRequestHandler():
         # Submit the event to the TANNER service
         event_result = await self.tanner_handler.submit_data(data)
 
-        if not event_result:
-            self.logger.error("Error submitting data: {}".format(data))
-
         # Log the event to slurp service if enabled
         if self.run_args.slurp_enabled:
             await self.submit_slurp(request.path_qs)
--- a/snare/server.py
+++ b/snare/server.py
@@ -49,6 +49,9 @@ class HttpRequestHandler():
         # Submit the event to the TANNER service
         event_result = await self.tanner_handler.submit_data(data)
 
+        if not event_result:
+            self.logger.error("Error submitting data: {}".format(data))
+
         # Log the event to slurp service if enabled
         if self.run_args.slurp_enabled:
             await self.submit_slurp(request.path_qs)
--- a/snare/tanner_handler.py
+++ b/snare/tanner_handler.py
@@ -58,7 +58,7 @@ class TannerHandler():
                     self.logger.error('Error submitting data: {} {}'.format(e, data))
                     event_result = {'version': '0.6.0', 'response': {'message': {'detection':
                                     {'name': 'index', 'order': 1, 'type': 1, 'version': '0.6.0'},
-                                    'sess_uuid': self.snare_uuid.decode('utf-8')}}}
+                                    'sess_uuid': data['uuid']}}}
                 finally:
                     await r.release()
         except Exception as e:
--- a/snare/tanner_handler.py
+++ b/snare/tanner_handler.py
@@ -54,7 +54,7 @@ class TannerHandler():
                 )
                 try:
                     event_result = await r.json()
-                except json.decoder.JSONDecodeError as e:
+                except (json.decoder.JSONDecodeError, aiohttp.client_exceptions.ContentTypeError) as e:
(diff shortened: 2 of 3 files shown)


afeena:
I just realize that just catching exception won't help, because here https://github.com/mushorg/snare/blob/master/snare/server.py#L56 we do not expect  `event_result` to be None. Would you mind add this check?

mzfr:
@afeena I've added a simple test to see if the complete dictionary is empty or not and then logged the error message. 
Is this what you were saying or should I check for that specific value i.e `event_result['response']` ?

afeena:
@mzfr my idea is to make tanner server error transparent for snare user. So in case of any tanner problem, the user will get the page without any injections. E.g. make default event result as `index` detection format with requested page :)

mzfr:
@afeena I've made the suggested changes.

afeena:
I got an error
```
Traceback (most recent call last):
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_protocol.py", line 390, in start
    resp = await self._request_handler(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_app.py", line 366, in _handle
    resp = await handler(request)
  File "/usr/local/lib64/python3.7/site-packages/aiohttp/web_middlewares.py", line 106, in impl
    return await handler(request)
  File "/usr/local/lib/python3.7/site-packages/Snare-0.3.0-py3.7.egg/snare/middlewares.py", line 22, in error_middleware
    response = await handler(request)
  File "/usr/local/lib/python3.7/site-packages/Snare-0.3.0-py3.7.egg/snare/server.py", line 74, in handle_request
    response_headers.add('Set-Cookie', 'sess_uuid=' + cur_sess_id)
TypeError: can only concatenate str (not "bytes") to str

```

remember, that `self.snare_uuid` has type bytes

mzfr:
yes and now the weird this is that if I do `self.snare_uuid.decode('utf-8')` the test fails because `str object has no attribute decode` but if we run it with docker/normally it just works without this error.

If I print those values first
```
print(self.snare_uuid)   # this should be with byte type
event_result = {'version': '0.6.0', 'response': {'message': {'detection':
                                    {'name': 'index', 'order': 1, 'type': 1, 'version': '0.6.0'},
                                    'sess_uuid': self.snare_uuid.decode('utf-8')}}}

print(event_result)
```
And here is the result:

image

We can see it's converted to a string. But pytest fails.

mzfr:
I just noticed one thing, in tanner_handler:submit_data function if we print the `data` values it gives:

```json
{
  "method": "GET",
  "path": "/",
  "headers": {
    "Host": "0.0.0.0",
    "User-Agent": "[agent]",
    "Accept": "[type]",
    "Accept-Language": "en-US,en;q=0.5",
    "Accept-Encoding": "gzip, deflate",
    "Connection": "keep-alive",
    "Cookie": "csrftoken=[token];sessionid=[sessionid];sess_uuid=922dab9f-b8d5-49db-a952-a3c469616036",
    "Upgrade-Insecure-Requests": "1"
  },
  "uuid": "8336fbc3-e635-45fe-9562-af0dce487539",
  "peer": {
    "ip": "172.20.0.1",
    "port": 59968
  },
  "status": 200,
  "cookies": {
    "csrftoken": "[token]",
    " sessionid": "[sessionid]",
    " sess_uuid": "922dab9f-b8d5-49db-a952-a3c469616036"
  }
}
```

and in the `test_tanner_handler_submit_data` in the setUp function we have `self.data` in the following way:

```json
{
  "method": "GET",
  "path": "/",
  "headers": {
    "sess_uuid": "test_uuid",
    "Accept": "text/html",
    "Accept-Encoding": "test_encoding",
    "Accept-Language": "test_lang",
    "Connection": "keep-alive",
    "Cookie": "test_cookie",
    "Host": "test_host",
    "Upgrade-Insecure-Requests": "1",
    "User-Agent": "test_agent",
    "cookies": "test_cookies",
    "peer": {
      "ip": "::1",
      "port": 80
    },
    "status": 200,
    "uuid": "test_uuid"
  }
}
```

mzfr:
@afeena I had to update the test case data for this. I've made all the required changes, please review and let me know if any other changes are required.

afeena:
@mzfr works well! :)

PR #249, opened 2020-02-29
Add option to delete a page directory

mzfr:
I thought it would be nice to add an option to delete an existing page directory.

This is how it would look:

```
➜ sudo python bin/snare --rm-pages example.com

   _____ _   _____    ____  ______
  / ___// | / /   |  / __ \/ ____/
  \__ \/  |/ / /| | / /_/ / __/
 ___/ / /|  / ___ |/ _, _/ /___
/____/_/ |_/_/  |_/_/ |_/_____/


Deleted page directory /opt/snare/pages/example.com
```

and if the directory doesn't exist then

```
➜ sudo python bin/snare --rm-pages example.com

   _____ _   _____    ____  ______
  / ___// | / /   |  / __ \/ ____/
  \__ \/  |/ / /| | / /_/ / __/
 ___/ / /|  / ___ |/ _, _/ /___
/____/_/ |_/_/  |_/_/ |_/_____/

    
Error: /opt/snare/pages/example.com - No such file or directory.
```

### DIFF
--- a/bin/snare
+++ b/bin/snare
@@ -21,6 +21,7 @@ import configparser
 import json
 import multiprocessing
 import os
+import shutil
 import sys
 import time
 import uuid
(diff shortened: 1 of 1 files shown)


afeena:
Could you please explain your idea? Why we need this?

mzfr:
I just thought that since have an option to clone the pages it would be nice to have option to remove those pages as well.

Instead of using ` sudo rm -rf /opt/pages/{name}`. There is nothing wrong in doing that but just thought it wouldn't hurt to have this option.

afeena:
1. I think running snare server only for page deletion is an overkill, I prefer `rm -rf`
2. Snare should not manage pages, it might be managed by cloner (or by another tool, I can image some "repository managing system" for pages, which includes not only cloner, but also some management system with such features as deletion) or by user.

mzfr:
oh okay so I think I should close this PR.

afeena:
I close it, because it need some more thinking around

PR #237, opened 2020-02-24
Fix KeyError

mzfr:
Fixes #201 

There was no `hash` key in the `/status_404`. Instead, that key is present in `/index.html`

```
{'/index.html': {'hash': 'd1546d731a9f30cc80127d57142a482b', 'content_type': 'text/html'}, '/status_404': {}}
```

So I fixed that and now snare starts properly.

PR #251, opened 2020-03-02
Add check for meta.json file

mzfr:
@afeena as you said in [this comment](https://github.com/mushorg/snare/pull/237#issuecomment-593109308) I've added a check for the `meta.json` file.

Let me know if any other changes are required.

### DIFF
--- a/bin/snare
+++ b/bin/snare
@@ -34,6 +34,26 @@ from snare.utils import snare_helpers
 from snare.utils.snare_helpers import str_to_bool, print_color
 
 
+def check_meta_file(meta_info):
+    """
+    This function checks the meta.json file
+
+    meta.json format is:
+    {'/page_path.html': {'hash':'long hash', 'content_type': 'page mime type'}}
+
+    This function makes sure none of the path in that is missing the `hash` value.
+
+    refer to issue: https://github.com/mushorg/snare/issues/201
+    """
+
+    for k, v in meta_info.items():
+        if 'hash' in v and 'content_type' in v:
+            continue
+        else:
+            return False
+    return True
+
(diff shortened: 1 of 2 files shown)


bin/snare
```diff
@@ -34,6 +34,26 @@ from snare.utils import snare_helpers
...
+def check_meta_file(meta_info):
+    """
+    This function checks the meta.json file
+
+    meta.json format is:
+    {'/page_path.html': {'hash':'long hash', 'content_type': 'page mime type'}}
+
+    This function makes sure none of the path in that is missing the `hash` value.
+
+    refer to issue: https://github.com/mushorg/snare/issues/201
```
afeena: No need for that

bin/snare
```diff
@@ -176,6 +196,11 @@ if __name__ == '__main__':
 
     with open(os.path.join(full_page_path, 'meta.json')) as meta:
         meta_info = json.load(meta)
+
+    if not check_meta_file(meta_info):
+        print_color("meta.json is missing some hash values. Please clone the pages again.", "ERROR")
```
afeena: Is it a good idea to point exact missing values?

bin/snare
```diff
@@ -34,6 +34,26 @@ from snare.utils import snare_helpers
 from snare.utils.snare_helpers import str_to_bool, print_color
 
 
+def check_meta_file(meta_info):
```
afeena: Maybe it's better to put it in utils? https://github.com/mushorg/snare/blob/master/snare/utils/snare_helpers.py

afeena:
Looks good, thanks!

PR #255, opened 2020-03-03
Printing progress for the pages being cloned

steve7158:
@afeena  As discussed earlier in slack i am printing the cloned urls as means for showing progress

### DIFF
--- a/bin/clone
+++ b/bin/clone
@@ -44,23 +44,24 @@ def main():
     else:
         log_file = "/opt/snare/clone.log"
     logger.Logger.create_clone_logger(log_file, __package__)
-    print("  Logs will be stored in {}".format(log_file), end="")
+    print_color("  Logs will be stored in {}".format(log_file),"INFO" ,end="")
     try:
         start=datetime.now()
         cloner = Cloner(args.target, int(args.max_depth), args.css_validate)
         loop.run_until_complete(cloner.get_root_host())
         loop.run_until_complete(cloner.run())
         end=datetime.now()-start
-        print("")
-        print("""
-        -------------------------------------->SUMMARY<------------------------------------------------
-         Total number of URLs cloned: {}
-         Time elapsed: {}
-         Cloned directory: {}
-        -----------------------------------------------------------------------------------------------
-        """.format(str(cloner.counter), str(end), cloner.target_path))
     except KeyboardInterrupt:
+        end=datetime.now()-start
         pass
+    finally:
+        print("")
+        print_color("-"*36+">SUMMARY<"+"-"*36,"INFO")
+        print_color('\tTotal number of URLs cloned: {}'.format(str(cloner.counter)),"INFO")
+        print_color('\tTime elapsed: {}'.format(str(end)),"INFO")
+        print_color('\tCloned directory: {}'.format(cloner.target_path),"INFO")
+        print_color('-'*82,"INFO")
+
 
 if __name__ == '__main__':
     print("""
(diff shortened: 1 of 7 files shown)


bin/clone
```diff
@@ -50,6 +50,7 @@ def main():
         cloner = Cloner(args.target, int(args.max_depth), args.css_validate)
         loop.run_until_complete(cloner.get_root_host())
         loop.run_until_complete(cloner.run())
+        print('Done!')
```
afeena: You can also print something like stats (e.g. cloned <N> pages in <M> minutes), and show the result directory

snare/cloner.py
```diff
@@ -140,6 +140,7 @@ def _make_filename(self, url):
             if data is not None:
                 self.meta[file_name]['hash'] = hash_name
                 self.meta[file_name]['content_type'] = content_type
+                print('Cloned url: ',current_url)
```
afeena: I'd prefer to keep stdout as clean as possible because if you have thousands of urls it will be super messy, I suggest you think about printing some counts or put some animation

afeena:
And please fix the travis build :) Thanks!

steve7158:
Thanks for the review @afeena .I'll get it done. :+1:

afeena:
@steve7158 Thanks! :)

bin/clone
```diff
@@ -45,15 +44,24 @@ def main():
...
+    print("  Logs will be stored in {}".format(log_file), end="")
     try:
+        start=datetime.now()
         cloner = Cloner(args.target, int(args.max_depth), args.css_validate)
         loop.run_until_complete(cloner.get_root_host())
         loop.run_until_complete(cloner.run())
+        end=datetime.now()-start
+        print("")
+        print("""
+        -------------------------------------->SUMMARY<------------------------------------------------
```
afeena: The easy and nice way to make multiple dashes is
`"-"*<N>` -> code looks nicer :)

bin/clone
```diff
@@ -45,15 +44,24 @@ def main():
     else:
         log_file = "/opt/snare/clone.log"
     logger.Logger.create_clone_logger(log_file, __package__)
-    print_color("Logs will be stored in {}\n".format(log_file), 'INFO')
+    print("  Logs will be stored in {}".format(log_file), end="")
```
afeena: why you need empty end here? and space in the beginning?

bin/clone
```diff
@@ -45,15 +44,24 @@ def main():
...
+        end=datetime.now()-start
+        print("")
+        print("""
+        -------------------------------------->SUMMARY<------------------------------------------------
+         Total number of URLs cloned: {}
+         Time elapsed: {}
+         Cloned directory: {}
+        -----------------------------------------------------------------------------------------------
+        """.format(str(cloner.counter), str(end), cloner.target_path))
     except KeyboardInterrupt:
```
afeena: what is we want summary also if we interrupt execution?

bin/clone
steve7158: Thanks for the advice i will get to it :smile: .

bin/clone
steve7158: This is for the animation -- 
image
Let me know if this looks ok


+ I will also add the print_color functionality to them as well.

bin/clone
steve7158: Agreed, I'll get it done :+1:

afeena:
Looks good :) resolve conflicts and I will merge

PR #264, opened 2020-03-16
Fixes KeyError, Index.html file not created

to09:
Sometimes while running snare, if  there is no index.html file, then it will raises the KeyError and program crashes.
It is not fixed.

### DIFF
--- a/bin/snare
+++ b/bin/snare
@@ -195,14 +195,14 @@ if __name__ == '__main__':
     if not check_meta_file(meta_info):
         print_color("Error found in meta.json. Please clone the pages again.", "ERROR")
         exit()
-
-    if not os.path.exists(os.path.join(base_page_path,
-                                       args.page_dir,
+    try:
+        if not os.path.exists(os.path.join(base_page_path, args.page_dir,
                                        os.path.join(meta_info[args.index_page]['hash']))):
-        print_color('can\'t create meta tag', 'WARNING')
-    else:
-        snare_helpers.add_meta_tag(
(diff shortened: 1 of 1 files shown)


afeena:
Hi @to09 , thanks! could you please give an example os such a case?

bin/snare
```diff
@@ -195,14 +195,14 @@ if __name__ == '__main__':
...
                                        os.path.join(meta_info[args.index_page]['hash']))):
-        print_color('can\'t create meta tag', 'WARNING')
-    else:
-        snare_helpers.add_meta_tag(
-            args.page_dir, meta_info[args.index_page]['hash'], config)
+            print_color('can\'t create meta tag', 'WARNING')
+        else:
+            snare_helpers.add_meta_tag(args.page_dir, meta_info[args.index_page]['hash'], config)
+    except KeyError as e:
+        print_color(e,"has not been created.")
```
afeena: this error message doesn't give me any clue how I should proceed, could we make it more informative?

to09: When I check all the "/opt/snare/pages" files created , I found one file created with "www.collegesolution.co.in" which raises an KeyError, 
I have attached screenshot for that 
Screenshot from 2020-03-17 17-05-54

afeena: Thanks for sharing! I meant that it would be nice to have more informative error message :)

afeena:
Could you please resolve conflicts?

to09:
I am trying to resolve this issue.

glaslos:
closing due to incativity.

PR #289, opened 2021-04-12
Fix assertion error for superuser

lordlabuckdas:
In [test_snare_helpers_check_privileges.py](https://github.com/mushorg/snare/blob/master/snare/tests/test_snare_helpers_check_privileges.py#L18), since a superuser would have (write) permissions for the root directory, **2** of the assertions **fail**.

So, I have used `os.geteuid()` to get the effective user ID and set `self.privileges` accordingly.

I have also removed the executable permissions for [snare_helpers.py](https://github.com/mushorg/snare/blob/master/snare/utils/snare_helpers.py) as it is not required.


glaslos:
I find this test a bit confusing. Are we testing here if we could run Snare on this system?

lordlabuckdas:
Yes, we're testing the `check_privileges` function which is used to create a directory for Snare and handle cases where the user does not have sufficient permissions to do so. I believe this makes Snare more non-root-user-friendly.
I agree that this test does not make much sense if Snare is intended to only be run as root.

PR #294, opened 2021-06-12
headless cloning using pyppeteer

lordlabuckdas:
added headless cloning using pyppeteer

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ beautifulsoup4==4.6.3
 cssutils==1.0.2
 gitpython==3.1.0
 pycodestyle==2.4.0
+pyppeteer==0.2.5
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -16,7 +16,7 @@ from pyppeteer import launch
 animation = "|/-\\"
 
 
-class Cloner(object):
+class BaseCloner:
     def __init__(self, root, max_depth, css_validate, default_path="/opt/snare"):
         self.logger = logging.getLogger(__name__)
         self.logger.setLevel(logging.DEBUG)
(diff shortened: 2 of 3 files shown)


snare/cloner.py
```diff
@@ -144,7 +153,64 @@ def _make_filename(self, url):
...
+        try:
+            driver.get(str(current_url))
+        except TimeoutException:
+            self.logger.error("Request timed out:" + str(current_url))
+        except InvalidArgumentException:
+            self.logger.error("Malformed URL:" + str(current_url))
+        except Exception as err:
+            self.logger.error(err)
+        time.sleep(1)
+        data = driver.page_source
```
afeena, line 198: do we want to stick to async version? All our functions are async, and I think it makes sense to use async libs like https://github.com/pyppeteer/pyppeteer
@lordlabuckdas @mzfr what do you think?

requirements.txt
```diff
@@ -1,6 +1,25 @@
 aiohttp==3.7.4
-aiohttp_jinja2==1.1.0
+aiohttp-jinja2==1.1.0
```
afeena, line 2: just out of curiosity, why so many new requirements? :) And do we need to set a version for each?

lordlabuckdas: Apart from the original `requirements.txt`, I only installed `pyppeteer`. Maybe these are dependencies? I think setting a version would make Snare more stable.

snare/cloner.py
```diff
@@ -155,16 +165,26 @@ async def get_body(self, session):
...
-            try:
-                response = await session.get(current_url, headers={"Accept": "text/html"}, timeout=10.0)
-                headers = self.get_headers(response)
-                content_type = response.content_type
-                data = await response.read()
-            except (aiohttp.ClientError, asyncio.TimeoutError) as client_error:
-                self.logger.error(client_error)
-            else:
-                await response.release()
+            if page:
```
afeena, line 170: Do you think it will be a good idea to split the headless/non-headless parts into the separate functions (or even classes with the common interface)?

PR #298, opened 2021-06-19
retrying of urls

lordlabuckdas:
* I have redesigned the `new_urls` queue to facilitate retrying of URLs.
* Each item of `new_urls` is a tuple of a tuple and integer - `((current_url: str, level: int), try_count: int)`.
* If the data is not successfully retrieved, the URL is again added to the queue with its`try_count` increased by one.
* The URL is tried for at-most 2 times before being dropped.
* Retrying of URLs is all the more important in the case of headless cloning where there is a high chance of timeouts or connection errors.

PR is currently in draft to accommodate changes of headless cloning.

### DIFF
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -95,7 +95,7 @@ class BaseCloner:
             ):
                 return None
         if url.human_repr() not in self.visited_urls and (level + 1) <= self.max_depth:
-            await self.new_urls.put((url, level + 1))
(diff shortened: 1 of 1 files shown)


afeena:
@lordlabuckdas could you please resolve the conflict?

lordlabuckdas:
Done :+1: @afeena

mzfr (approved):
changes looks good to me 👍

PR #302, opened 2021-07-13
feature: redirects

lordlabuckdas:
### Major changes

Pages can now redirect, similar to the actual website.

`meta.json` structure:

```json
{
    "/":
        {
            "redirect": "/new/home/"
        },
    "/new/home/":
        {
            "hash": "fb4b89f45c474f455a089130db94b00a",
            "headers": [{"Content-Type": "text/html"}]
        }
}
```

### Minor changes

* Replaced `counter` with `len(visited_urls)`
* `Transfer-Encoding` header has been added to the ignore list

Issue #230, opened 2020-01-31
KeyError: '/status_404'


MaheswarReddy1194321:
![key_error](https://user-images.githubusercontent.com/48216197/73519912-4d086d80-4428-11ea-9244-ea15d15470e4.png)

-> it's giving error if we are cloning website which don't have error page.

afeena:
@MaheswarReddy1194321 sorry for the late reply

I can't reproduce your problem :( But I know we had some problems with a cloner, so it doesn't work stable all the time. You should check you meta.json and try probably to re-clone again

afeena:
Maybe it's somehow related https://github.com/mushorg/snare/issues/215

afeena:
And this https://github.com/mushorg/snare/issues/183

lordlabuckdas:
As per my observation, this issue arises in 2 cases:
1. The target website **does not have an error page**.
2. Upon visiting a page that does not exist, we are **redirected** to the homepage.

The website in question (used by @MaheswarReddy1194321) comes under the 2nd case as `in.yahoo.com/status_404` redirects to `in.yahoo.com/?err=404&err_url=https%3a%2f%2fin.yahoo.com%2fstatus_404`.

PR #303, opened 2021-07-16
fix missing error page and improper redirects

lordlabuckdas:
As encountered in #230. some websites do not have a 404 template configured or redirect to the home page. To fix, this
1. An empty page is created
2. Or if the website redirects, it is caught by comparing the requested and returned URL (ref #302).

**Other fixes:**

* The `get_headers(response_headers: CIMultiDict) -> headers: list` method now returns the content type as well. This was done to optimize the working of Cloner and prevent a call to `get_content_type` which iterates through the list of headers to fetch the content type. From
    `get_headers(response_headers: CIMultiDict) -> headers: list` to
    `get_headers(response_headers:CIMultiDict) -> list[headers: list, content_type: str]`

* Moved the `Content-Type` header from `headers` key in meta info to a separate key
    From
    ```json
    "/about": {
        "hash": "abcd",
        "headers": [{"Server": "Apache"}, {"Content-Type": "text/html"}]
    }
    ```
    to
    ```json
    "/about": {
        "hash": "abcd",
        "headers": [{"Server": "Apache"}],
        "content_type": "text/html"
    }
    ```
    This was done because [tanner_handler.py#L111](https://github.com/mushorg/snare/blob/c917c4b2082ef5c1cea9a3236939b3a6a4015452/snare/utils/snare_helpers.py#L111) and [snare_helpers.py#L96](https://github.com/mushorg/snare/blob/c917c4b2082ef5c1cea9a3236939b3a6a4015452/snare/tanner_handler.py#L96) already check for it and overwrite the `Content-Type` header.

### DIFF
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -294,13 +294,6 @@ class CloneRunner:
     async def close(self):
         if not self.runner:
             raise Exception("Error initializing cloner!")
-        error_file_name, error_file_hash = self.runner._make_filename(self.runner.error_page)
-        # create empty file for 404 page if not present and add meta info
-        if not self.runner.meta.get(error_file_name):
-            with open(os.path.join(self.runner.target_path, error_file_hash), "wb") as _:
-                pass
-            self.runner.meta[error_file_name]["hash"] = error_file_hash
-            self.runner.meta[error_file_name]["content_type"] = "text/html"
         with open(os.path.join(self.runner.target_path, "meta.json"), "w") as mj:
             json.dump(self.runner.meta, mj)
         if self.driver:
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -68,11 +68,14 @@ class BaseCloner:
             "x-cache",
         ]
 
+        content_type = None
         headers = []
         for key, value in response.headers.items():
-            if key.lower() not in ignored_headers_lowercase:
+            if key.lower() == "content-type":
+                content_type = value
+            elif key.lower() not in ignored_headers_lowercase:
                 headers.append({key: value})
-        return headers
+        return [headers, content_type]
 
     async def process_link(self, url, level, check_host=False):
         try:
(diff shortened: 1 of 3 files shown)


snare/cloner.py
```diff
@@ -299,6 +294,13 @@ async def run(self):
     async def close(self):
         if not self.runner:
             raise Exception("Error initializing cloner!")
```
afeena, line 296: I just spot we have initialization error in close function 🤔

snare/cloner.py
```diff
@@ -299,6 +294,13 @@ async def run(self):
...
+        error_file_name, error_file_hash = self.runner._make_filename(self.runner.error_page)
```
afeena, line 297: Do we really need an empty page? Why not letting aiohttp handle this?
https://docs.aiohttp.org/en/latest/web_exceptions.html

snare/cloner.py
lordlabuckdas: Strangely enough, that check is there just to be thorough and satisfy linter warnings. If you are referring to an instance where the error was actually raised, could you please share the steps to replicate it? I've tried it with multiple websites and it seems to work fine.

snare/cloner.py
mzfr: Yeah I agree instead of generating a new empty page we should just let aiohttp handle that

snare/server.py
```diff
@@ -71,18 +69,22 @@ async def handle_request(self, request):
             if previous_sess_uuid is None or not previous_sess_uuid.strip() or previous_sess_uuid != cur_sess_id:
                 headers.add("Set-Cookie", "sess_uuid=" + cur_sess_id)
 
+        if status_code == 404:
+            raise web.HTTPNotFound(headers=headers)
+
         return web.Response(body=content, status=status_code, headers=headers)
```
lordlabuckdas, line 75: Special if-check for status code 404 because returning `web.Response` with `status_code=404` did not work as intended.

PR #304, opened 2021-07-20
updated tests

lordlabuckdas:
* Added and updated old tests to be compatible with the new cloner architecture.
* Some mocks (like that of `aiohttp.ClientResponse.read` and `aiohttp.ClientResponse._headers`) influenced tests other than their own and that caused some tests to fail. To prevent this, I have used `unittest.mock.patch` by assigning it to a variable (or appending it to a list if there are many mocks) to track (start and stop) them.
* Formatted code and organized imports (ref last commit).


snare/cloner.py
```diff
@@ -207,10 +212,10 @@ async def get_body(self, driver):
     async def get_root_host(self):
         try:
             async with aiohttp.ClientSession() as session:
-                async with session.get(self.root) as resp:
-                    resp_url = yarl.URL(resp.url)
-                    if resp_url.host != self.root.host or resp_url.path != self.root.path:
-                        self.moved_root = resp_url
+                resp = await session.get(self.root)
```
afeena, line 215: why have you decided to remove async context manager?

afeena:
@lordlabuckdas could you please resolve conflicts?

snare/tests/test_cloner_add_scheme.py
```diff
@@ -16,22 +18,28 @@ def setUp(self):
         self.expected_err_url = yarl.URL("http://example.com/status_404")
         self.max_depth = sys.maxsize
         self.css_validate = False
-        self.handler = Cloner(self.url, self.max_depth, self.css_validate)
+        self.handler = BaseCloner(self.url, self.max_depth, self.css_validate)
 
     def test_trailing_slash(self):
         self.url = "http://example.com/"
+        if not self.handler:
```
afeena, line 25: you have this check in multiple tests, why not checking it in setup?

snare/tests/test_cloner_fetch_data.py
```diff
@@ -0,0 +1,96 @@
...
+        self.headless_handler = HeadlessCloner(self.url, self.max_depth, self.css_validate)
+
+    def test_basecloner_fetch_data(self):
+        async def test():
+            with self.assertRaises(NotImplementedError):
+                await self.base_handler.fetch_data(None, self.url, 0, 0)
+
+        self.loop.run_until_complete(test())
+
+    def test_simplecloner_fetch_data(self):
```
afeena, line 39: what are you testing here? what is the purpose of this test? what is expected behaviour? Right now it looks for me not like an unittest test, but rather like just running the code

snare/tests/test_cloner_fetch_data.py
```diff
@@ -0,0 +1,96 @@
...
+            try:
+                session = aiohttp.ClientSession()
+                await self.simple_handler.fetch_data(session, self.url, 0, 0)
+            finally:
+                if session:
+                    await session.close()
+
+        self.loop.run_until_complete(test())
+
+    def test_headlesscloner_fetch_data(self):
```
afeena, line 51: see my previous comment

snare/tests/test_cloner_fetch_data.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_headlesscloner_fetch_data_redirect(self):
+        self.expected_redirect_url = yarl.URL("http://www.example.com")
+        p = patch("pyppeteer.network_manager.Response.url", new=self.expected_redirect_url)
+        p.start()
+
+        self.redirect_url = None
+
+        async def test():
+            browser = None
+            try:
```
afeena, line 83: test should check how function handles the exception, it should not handle it

snare/tests/test_cloner_get_body.py
```diff
@@ -48,113 +51,228 @@ def setUp(self):
...
+            patch(
+                "aiohttp.ClientResponse.read",
+                new=AsyncMock(side_effect=[self.index_page_content, self.test_page_content]),
+            )
+        )
+        for p in patches:
+            p.start()
+
+        if not self.handler.runner:
+            raise Exception("Error initializing Cloner!")
```
afeena, line 96: see previous comments

snare/tests/test_cloner_get_body.py
```diff
@@ -48,113 +51,228 @@ def setUp(self):
...
         self.content = b""".banner { background: url("/example.png") }"""
-        aiohttp.ClientResponse.read = AsyncMock(return_value=self.content)
+
+        patches = []
+        patches.append(patch("aiohttp.ClientResponse._headers", new={"Content-Type": "text/css", "Server": "Apache"}))
+        patches.append(patch("aiohttp.ClientResponse.read", new=AsyncMock(return_value=self.content)))
+        for p in patches:
+            p.start()
+
+        self.session.get = AsyncMock(
```
afeena, line 148: you have similar mock before, can it be reused?

snare/tests/test_cloner_init.py
```diff
@@ -1,18 +1,45 @@
...
 class TestClonerInitialization(unittest.TestCase):
     def setUp(self):
         self.root = "http://example.com"
         self.max_depth = sys.maxsize
         self.css_validate = False
-        self.handler = Cloner(self.root, self.max_depth, self.css_validate, default_path="/tmp")
+        self.target_path = None
+        self.handler = None
+
+    def test_base_cloner_init(self):
```
afeena, line 17: seems rather useless

afeena: do we really need these init tests? what it brings? how it helps to prevent errors?

snare/tests/test_cloner_init.py
```diff
@@ -1,18 +1,45 @@
...
+        self.handler = BaseCloner(self.root, self.max_depth, self.css_validate, default_path="/tmp")
+        self.target_path = self.handler.target_path
+        self.assertIsInstance(self.handler, BaseCloner)
+
+    def test_simple_cloner_init(self):
```
afeena, line 22: see above

snare/tests/test_cloner_init.py
```diff
@@ -1,18 +1,45 @@
...
+        self.handler = SimpleCloner(self.root, self.max_depth, self.css_validate, default_path="/tmp")
+        self.assertIsInstance(self.handler, SimpleCloner)
+
+    def test_headless_cloner_init(self):
```
afeena, line 26: same

afeena (changes requested):

afeena:
A lot of code repetition, I think a lot of tests can be grouped to reuse the same inits etc

snare/tests/test_cloner_add_scheme.py
lordlabuckdas: I had kept it for linter checks, but with my recent push I have now shifted it to `setUp` as you suggested.

snare/tests/test_cloner_fetch_data.py
lordlabuckdas: I have now removed some of the tests from `fetch_data` as the function itself is pretty straight-forward and the tests for `get_body` already cover its general use-case.

snare/tests/test_cloner_fetch_data.py
lordlabuckdas: The `try-finally` block was not intended to handle the errors, but I realized it was not very useful and have now replaced it with the direct calls.

snare/tests/test_cloner_get_body.py
lordlabuckdas: I think each of the mocks are unique, so they can't be reduced, but to reduce the visual clutter, I have introduced a separate `create_client_response` method which takes the URL and event loop as arguments to return the ClientResponse object.

snare/tests/test_cloner_init.py
lordlabuckdas: I have now kept only the test where there is a initialization error and have removed the rest. On this note, what do you think about the tests for the run method?

PR #305, opened 2021-07-23
accurate exception handling in page.close

lordlabuckdas:
* If `KeyboardInterrupt` is raised midway headless cloning, then a `PageError` is raised while closing the `Page` object. The raised exception can be safely ignored as the execution comes to a halt after it but there is no way to avoid a try-except in the finally part of `HeadlessCloner.fetch_data`.
* The `Content-Type` header might sometimes contain character set information like, `Content-Type: text/html; charset=utf-8`. To accommodate this, we can check for *presence* rather than *equality*. Sometimes, `Content-Type` is not set and to avoid `TypeError` by iteration in a `NoneType` object, we check if it is not `None` first.

### DIFF
--- a/snare/cloner.py
+++ b/snare/cloner.py
@@ -183,10 +183,12 @@ class BaseCloner:
             self.meta[file_name]["headers"] = headers
             self.meta[file_name]["content_type"] = content_type
 
-            if content_type == "text/html":
+            if not content_type:
+                pass
+            elif "text/html" in content_type:
                 soup = await self.replace_links(data, level)
                 data = str(soup).encode()
-            elif content_type == "text/css":
+            elif "text/css" in content_type:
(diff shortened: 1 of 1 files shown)


snare/cloner.py
```diff
@@ -182,10 +183,12 @@ async def get_body(self, driver):
             self.meta[file_name]["headers"] = headers
             self.meta[file_name]["content_type"] = content_type
 
-            if content_type == "text/html":
+            if not content_type:
```
mzfr, line 186: I think writing this block in the following way would be nice

```python
if content_type:
	if "text/html" in content_type:
		.....
	elif/else ANOTHER_CONDITION_HERE:
		.....
```

Other than this everything looks good :)

mzfr (approved):

PR #308, opened 2021-08-05
error templating fixes

lordlabuckdas:
* Snare cannot be made completely non-fingerprint-able because a malformed request triggers a 400 exception which cannot be handled by aiohttp _at the moment_ (ref [aiohttp#5358](https://github.com/aio-libs/aiohttp/issues/5358) and [aiohttp#3287](https://github.com/aio-libs/aiohttp/issues/3287)).
* This PR ensures that proper headers are assigned during 404 and 500 errors.
* `HttpRequestHandler.remove_default_server_header` removes the default Server header to prevent leakage of the aiohttp server banner.

### DIFF
--- a/snare/middlewares.py
+++ b/snare/middlewares.py
@@ -4,9 +4,8 @@ import multidict
 
 
 class SnareMiddleware:
-    def __init__(self, error_404, error_500=None, headers=[], server_header=""):
+    def __init__(self, error_404, headers=[], server_header=""):
         self.error_404 = error_404
-        self.error_500 = error_500 if error_500 else "500.html"
 
         self.headers = multidict.CIMultiDict()
         for header in headers:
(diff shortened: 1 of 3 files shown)


snare/middlewares.py
```diff
@@ -17,10 +16,15 @@ def __init__(self, error_404, error_500=None, headers=[], server_header=""):
             self.headers["Server"] = server_header
 
     async def handle_404(self, request):
-        return aiohttp_jinja2.render_template(self.error_404, request, {})
-
-    async def handle_500(self, request):
-        return aiohttp_jinja2.render_template(self.error_500, request, {})
+        if not self.error_404:
```
mzfr, line 19: @lordlabuckdas Can you please explain what is happening here? Why are we triggering NotFound error when `self.error_404` is not set? Shouldn't it be triggered when `self.error_404` is set?

lordlabuckdas: `self.error_404` contains the hash/file name of the 404 page. In case websites do not have a 404 template set, `self.error_404` should be `None` and so we're triggering the exception manually with the headers. If the file is present, we're serving the page using Jinja's `render_template` function.

mzfr: Oh okay, I thought a boolean value would be assigned to the error_404 variable.

Your explanation makes sense.

PR #309, opened 2021-08-15
add docstrings

lordlabuckdas:


snare/cloner.py
```diff
@@ -21,7 +22,18 @@
 
 
 class BaseCloner:
-    def __init__(self, root, max_depth, css_validate, default_path="/opt/snare"):
+    def __init__(self, root: str, max_depth: int, css_validate: bool, default_path: str = "/opt/snare") -> None:
+        """Base class for all core functions of the cloner
+
+        :param root: Website root URL
+        :type root: str
```
afeena, line 29: do we need type if the type annotated in function definition?

snare/utils/snare_helpers.py
```diff
@@ -33,11 +40,17 @@ def check_compatibility(self, tanner_version):
...
 class Converter:
-    def __init__(self):
+    def __init__(self) -> None:
+        """Converter class"""
         self.logger = logging.getLogger(__name__)
         self.meta = {}
 
-    def convert(self, path):
+    def convert(self, path: str) -> None:
+        """Convert all pages to a Snare-friendly form and write meta info
```
afeena, line 49: what is Snare-friendly format? :)

snare/cloner.py
```diff
@@ -111,7 +151,16 @@ async def process_link(self, url, level, check_host=False):
             self.logger.error("ValueError while processing the %s link", url)
         return res
 
-    async def replace_links(self, data, level):
+    async def replace_links(self, data: Union[bytes, str], level: int) -> BeautifulSoup:
+        """Replace website links to make them suitable for serving
```
afeena, line 155: what does it mean "suitable"? try to be more concrete, maybe even with example

snare/cloner.py
```diff
@@ -242,11 +331,26 @@ async def fetch_data(self, session, current_url, level, try_count):
             await self.new_urls.put({"url": current_url, "level": level, "try_count": try_count + 1})
         else:
             await response.release()
-        return [redirect_url, data, headers, content_type]
+        return redirect_url, data, headers, content_type
 
 
 class HeadlessCloner(BaseCloner):
-    async def fetch_data(self, browser, current_url, level, try_count):
+    async def fetch_data(
```
afeena, line 338: consider adding the docstring for the class as well

snare/cloner.py
lordlabuckdas: Yes, Sphinx [mandates it](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format) and the type disappears from the parameter description in docs when we remove its `:type` from the docstring.

snare/cloner.py
lordlabuckdas: Have clarified it in 59dd4d2. Will add detailed descriptions to this and other functions in the next series of commits.

snare/utils/snare_helpers.py
lordlabuckdas: I meant to convey the renaming of files to their hashes but I realize it sounds very vague. I have clarified it in 59dd4d2.

PR #328, opened 2023-12-09
Update snare

hrchlhck:
I've been facing an error when I tried to run Snare. The environment was a Docker Container consisting of an Ubuntu image, version 20.04.

The error I encountered was "The port is required to be int." even using the flag "--port".

I've also tried to change the Docker image version to 22.04, and updated the `aiohttp` library to the most updated version. Even though, I got no success unless changing the type of the argument.


afeena:
Thank you for your contribution!

